PR #12, opened 2020-05-07
172411943 measurement class

simonwa7:
Creating a class for `Measurements` and reconciling overlap with `BitstringDistribution` class. 

This PR also affects the `QuantumSimulator` interface and therefore resources implementing this interface may need to be updated

### DIFF
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from ..bitstring_distribution import BitstringDistribution, create_bitstring_distribution_from_probability_distribution, create_bitstring_distribution_from_measurements
+from ..bitstring_distribution import BitstringDistribution, create_bitstring_distribution_from_probability_distribution
 from ..circuit import Circuit
 from ..measurement import ExpectationValues
 from typing import Optional, List, Tuple
@@ -23,6 +23,8 @@ class QuantumBackend(ABC):
         Method for executing the circuit and measuring the outcome.
         Args:
             circuit (core.circuit.Circuit): quantum circuit to be executed.
+        Returns:
+            Measurements: object representing the measurements resulting from the circuit
         """
         raise NotImplementedError
 
@@ -92,4 +94,4 @@ class QuantumSimulator(QuantumBackend):
         else:
             # Get the expectation values
             measurements = self.run_circuit_and_measure(circuit, **kwargs)
-            return create_bitstring_distribution_from_measurements(measurements)
+            return measurements.get_distribution()
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -24,7 +24,7 @@ class QuantumBackend(ABC):
         Args:
             circuit (core.circuit.Circuit): quantum circuit to be executed.
         Returns:
-            Measurements: object representing the measurements resulting from the circuit
+            core.measurement.Measurements: object representing the measurements resulting from the circuit
         """
         raise NotImplementedError
 
--- a/src/python/zquantum/core/measurement.py
+++ b/src/python/zquantum/core/measurement.py
@@ -397,7 +397,7 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
 class Measurements:
     """ A class representing measurements from a quantum circuit. The bitstrings variable represents the internal
     data structure of the Measurements class. It is expressed as a list of tuples wherein each tuple is a measurement
-    and the value of the tuple at a given index is the measured bit-value of the qubit (indexed from N-1 -> 0) """
+    and the value of the tuple at a given index is the measured bit-value of the qubit (indexed from 0 -> N-1) """
 
     def __init__(self, bitstrings: Optional[List[Tuple[int]]] = None):
         if bitstrings is None:
@@ -407,7 +407,7 @@ class Measurements:
 
 
     @classmethod
-    def from_counts(cls, counts: Dict):
+    def from_counts(cls, counts: Dict[str, int]):
         """ Create an instance of the Measurements class from a dictionary
         
         Args:
@@ -416,6 +416,7 @@ class Measurements:
         measurements = cls()
         measurements.add_counts(counts)
         return measurements
+
 
     @classmethod
     def load_from_file(cls, file: TextIO):
@@ -460,11 +461,13 @@ class Measurements:
         return dict(Counter(bitstrings))
 
 
-    def add_counts(self, counts: Dict):
+    def add_counts(self, counts: Dict[str, int]):
         """ Add measurements from a histogram
 
         Args:
             counts (dict): mapping of bitstrings to integers representing the number of times the bitstring was measured
+                NOTE: bitstrings are also indexed from 0 -> N-1, where the "001" bitstring represents a measurement of 
+                    qubit 2 in the 1 state
         """
         for bitstring in counts.keys():
             measurement = []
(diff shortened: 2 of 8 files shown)


simonwa7:
This PR also fixes versioning of packages that have had breaking changes (`cirq` and `qiskit`)

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -23,6 +23,8 @@ def run_circuit_and_measure(self, circuit:Circuit, **kwargs) -> List[Tuple]:
         Method for executing the circuit and measuring the outcome.
         Args:
             circuit (core.circuit.Circuit): quantum circuit to be executed.
+        Returns:
+            Measurements: object representing the measurements resulting from the circuit
```
mstechly, line 27: Actually it should be named "core.measurements.Measurements" I guess ;)

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+            filename (string): filename to save the data to 
+        """
+        data = { "schema":         SCHEMA_VERSION+"-measurements",
+                 "counts":         self.get_counts(),
+                 "measurements":   self.get_measurements()}
+        with open(filename, "w") as f:
+            f.write(json.dumps(data, indent=2))
+
+
+    def get_measurements(self):
```
mstechly, line 424: Why do you need a getter for that?
The fields in the method are not private, so you can just do:
`measurements.measurements`, not need for `measurements.get_measurements()`.
You now have two ways to do the same thing.

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
             significant (little endian).
     """
     return int("".join(str(bit) for bit in bitstring[::-1]), 2)
+
+class Measurements:
+    """ A class representing measurements from a quantum circuit. """
+
+    def __init__(self, filename=None):
+        self.counts = {}
+        self.measurements = []
```
mstechly, line 402: Actually, perhaps "bitstrings" is better?
Cause measurements.measurements looks a little bit confusing to me ;)

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+    def add_measurements(self, measurements):
+        """ Add a list of measurement tuples 
+
+        Args:
+            measurements (list of tuples): A list of the measurements to add
+        """
+        for measurement in measurements:
+            self.measurements.append(tuple(measurement))
+
+            # Bitstrings are in reversed-index order compared to measurement tuples
```
mstechly, line 460: Why?
I don't think we should have such inconsistencies in the core library.

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+        return self.num_measurements
+
+
+    def add_measurements(self, measurements):
+        """ Add a list of measurement tuples 
+
+        Args:
+            measurements (list of tuples): A list of the measurements to add
+        """
+        for measurement in measurements:
```
mstechly, line 457: This for loop (and the following ones as well) look like potential performance bottleneck.
Isn't there a more pythonic way to do that?

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+                bitstring += str(bit_value)
+            
+            if bitstring in self.counts.keys():
+                self.counts[bitstring] += 1
+            else:
+                self.counts[bitstring] = 1
+            self.num_measurements += 1
+
+
+    def add_counts(self, counts):
```
mstechly, line 472: Why do we need to pass `counts` as an argument?
Shouldn't it calculate counts directly from `self.measurements`?

src/python/zquantum/core/interfaces/backend.py
simonwa7: Addressed here: 
5309e30

src/python/zquantum/core/measurement.py
simonwa7: Addressed here: 
9e00518

src/python/zquantum/core/measurement.py
simonwa7: This is to allow the user to add measurements to the object in either form. I believe many simulators (qiskit for example) return similar objects to the `counts` dictionary, so this avoids them having to create lists of tuples when they really just want to add the `counts`.

src/python/zquantum/core/measurement.py
simonwa7: Addressed in refactor here: 
231a59a

src/python/zquantum/core/measurement.py
simonwa7: Addressed in refactor here:
231a59a

src/python/zquantum/core/measurement.py
simonwa7: Addressed in refactor here:
231a59a

simonwa7:
Since we chose to order measurement tuples in reverse index order, we should make sure measurements are properly indexed when we propagate these changes to the resources implementing the QuantumSimulator interface

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,96 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+        """ Get the measurements as a histogram
+
+        Returns:
+            A dictionary mapping bitstrings to integers representing the number of times the bitstring was measured
+        """
+        bitstrings = convert_tuples_to_bitstrings(self.bitstrings)
+        return dict(Counter(bitstrings))
+
+
+    def add_counts(self, counts: Dict):
```
mstechly, line 461: FYI you an specify what should be the form of the Dict – e.g. Dict[Tuple[int], int] .
Not necessary to do so though.

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,96 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+        else:
+            data = json.load(file)
+
+        bitstrings = []
+        for bitstring in data["bitstrings"]:
+            bitstrings.append(tuple(bitstring))
+
+        return cls(bitstrings=bitstrings)
+
+    def save(self, filename: String):
```
mstechly, line 438: I think it should be `str` not `String` – the typing requires those funny capital letter names only for the "special type" or "collection type" of variables.
See: https://docs.python.org/3/library/typing.html

src/python/zquantum/core/measurement.py
```diff
@@ -392,3 +393,115 @@ def convert_bitstring_to_int(bitstring: Iterable[int]) -> int:
...
+
+class Measurements:
+    """ A class representing measurements from a quantum circuit. """
+
+    def __init__(self, filename=None):
+        self.counts = {}
+        self.bitstrings = []
+        self.num_measurements = 0
+
+        if filename is not None:
```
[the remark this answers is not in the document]
simonwa7: Already addressed in previous commits

src/python/zquantum/core/measurement.py
simonwa7: Good to know!

src/python/zquantum/core/measurement.py
simonwa7: Addressed here: 
7288dca

PR #18, opened 2020-05-14
172747567 cost function

mstechly:


jgonthier:
Hey @simonwa7 ! I'm weighting in as I was discussing with @mstechly the design of these classes. The idea was that a numerical gradient implementation can be done directly in the base class, as numerical gradients formulae are the same for all functions. But then we also want the user to be able to check their analytical gradient implementation against the numerical one, which I think is why there is an option to specify.

I guess an alternative would be to let the user define these options in their derived classes, so that the base method always does numerical gradients, and then the derived class can choose between user-implemented analytical formulae and the base class implementation of numerical gradients.

But maybe there's a better way, the ideas above are more C++ than Python I think.

mstechly:
@simonwa7 @jgonthier 
Actually, right now there is not distinction between `numerical` an `analytical` gradient in the base class, there's just `get_gradient`. Such distinction exists only for the `BasicCostFunction`, where I think it actually makes sense.

Regarding having a "numerical gradient" in the base class as Jerome said, I had it there, but decided to remove it for now. We can always add it in future, but I wasn't 100% sure if that's the right decision.

As we've discussed yesterday, I think that instead of the field `self.use_analytical_gradient` we could just have `self.gradient_type`, which gives more flexibility and leaves out the ambiguity of what `analytical` actually means`. It will also be easier to connect that with ansatz interface, since implementation of `get_gradient` could be in some cases just one line: `self.ansatz.get_gradient(params, self.gradient_type)` and the logic of checking whether given gradient is implemented will lie on the end of ansatz.

Anyway – if you approve, I'll change `self.use_analytical_gradient` to `self.gradient_type`, refactor code to reflect this change and leave the rest as it is.

simonwa7:
Okay, that makes sense to me. Then I would propose the following (in-line with @mstechly above):

1. Remove the `use_analytical_gradients` input parameter in the interface
2. Rename `get_numerical_gradients` to `get_gradients_finite_difference`
3. [Optional] Move `get_gradients_finite_difference` to the interface since it should be the same for any cost function and then it would allow us to make the default behavior of the `get_gradients` method to use `get_gradients_finite_difference`
4. Leave the logic for switching between gradient methods up to the classes themselves, yet establish a pattern we think is appropriate (ie using `self.gradient_type` to switch between different methods)
5. Have `BasicCostFunction` have `self.custom_gradient` which is a `Callable` and, when set, the behavior of `get_gradients` is to call `self.custom_gradient` and when not set, the behavior uses `self.get_gradients_finite_difference`

simonwa7:
Additionally, I'm not sure about the `EvaluateOperatorCostFunction` taking in the ansatz and backend upon initialization. To me, it seems that they should be passed as arguments to the `evaluate` method. I think it's conceivable that one would want to change the ansatz and/or backend without conceptually changing the cost function (in practice, I know you could just reassign the variables in the object, but that's counterintuitive to when an object is initialized with certain values).

mstechly:
@simonwa7 The rationale behind taking it upon initialization is that for the optimizer it should be just a box with `evaluate(params)` method. 
The optimizer should not care whether it minimizes something that requires an ansatz, mathematical formula or calling an endpoint at the other end of the Internet. It just cares about being able to run `evaluate(params)` method, so you can use any optimizer with any cost function.

If you want, you can create two copies of the cost function, where you use each with different backend. Also, given Orquestra architecture, you probably will do that in separate tasks anyway.

src/python/zquantum/core/cost_function.py
```diff
@@ -0,0 +1,144 @@
...
+        evaluations_history (list): List of the tuples (parameters, value) representing all the evaluation in a chronological order.
+        gradient_type (str): see Args
+        save_evaluation_history (bool): see Args
+        epsilon (float): see Args
+
+    """
+
+    def __init__(self, function:Callable, 
+                        gradient_function:Optional[Callable]=None, 
+                        gradient_type:str='custom',
```
simonwa7, line 33: How come we are defaulting to `custom` instead of `finite_difference`?

src/python/zquantum/core/cost_function.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class EvaluateOperatorCostFunction(CostFunction):
+    """
+    Cost function used for evaluating given operator using given ansatz.
+
+    Args:
+        target_operator (openfermion.QubitOperator): operator to be evaluated
+        ansatz (dict): dictionary representing the ansatz
+        backend (zquantu.core.interfaces.backend.QuantumBackend): backend used for evaluation
```
simonwa7, line 86: `zquantum` not `zquantu`

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -20,12 +21,12 @@ def __init__(self, options:Optional[Dict]=None):
...
 
     @abstractmethod
-    def minimize(self, cost_function:Callable, initial_params:np.ndarray, **kwargs) -> OptimizeResult:
+    def minimize(self, cost_function:CostFunction, initial_params:np.ndarray, **kwargs) -> OptimizeResult:
         """
         Finds the parameters which minimize given cost function.
 
         Args:
-            cost_function: a cost function to be minimized, depends on some numerical parameters.
+            cost_function (zquantu.core.interfaces.CostFunction): an object representing the cost function.
```
simonwa7, line 29: `zquantum` not `zquantu`

src/python/zquantum/core/utils.py
```diff
@@ -251,10 +251,14 @@ def to_dict(self):
...
-        data['value'] = self.value
+        if type(self.value).__module__ == np.__name__:
+            data['value'] = self.value.item()
+        else:
+            data['value'] = self.value
 
-        if self.precision:
+        if type(self.precision).__module__ == np.__name__:
+            data['precision'] = self.precision.item()
+        else:
```
simonwa7, line 261: Can you explain what's going on here?

src/python/zquantum/core/cost_function.py
mstechly: If user provides the `gradient_function`, they would expect this function to be used. So we default to `custom` and fall back to `finite_difference` if `custom` is not provided.
I think this way we'll cause less harm if someone mess it up.

src/python/zquantum/core/cost_function.py
mstechly: Fixed in `151e0a`

src/python/zquantum/core/interfaces/optimizer.py
mstechly: Fixed in `151e0a`

src/python/zquantum/core/utils.py
mstechly: This is for casting from `np.float64` to python `float`. 
Cause `np.float64` is non-serializable so it caused problems.
I agree this doesn't look well, but it's the best way I found.

PR #20, opened 2020-05-15
Fixing error when calculating number of qubits in probability distrib…

simonwa7:
…ution. Added test

The number of qubits was incorrectly being calculated as the square root of the length of the probability distribution instead of the logarithm. This PR fixes said bug and introduces a unittest for the given case.


simonwa7:
@martamau

PR #21, opened 2020-05-20
172420671 common tests

simonwa7:
This PR makes a base set of tests that all classes implementing the `QuantumBackend` or `QuantumSimulator` interfaces need to satisfy.

### DIFF
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -41,7 +41,7 @@ class QuantumBackend(ABC):
         """
         raise NotImplementedError
 
-    def get_expectation_values_for_circuitset(self, circuitset:[Circuit], qubit_operator:QubitOperator, **kwargs) -> [ExpectationValues]:
+    def get_expectation_values_for_circuitset(self, circuitset:List[Circuit], qubit_operator:QubitOperator, **kwargs) -> [ExpectationValues]:
         """
         Calculates the expectation values for given operator, based on the exact quantum state 
         produced by a set of circuits.
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -41,6 +41,38 @@ class QuantumBackend(ABC):
         """
         raise NotImplementedError
 
+    def get_expectation_values_for_circuitset(self, circuitset:[Circuit], qubit_operator:QubitOperator, **kwargs) -> [ExpectationValues]:
+        """
+        Calculates the expectation values for given operator, based on the exact quantum state 
+        produced by a set of circuits.
+
+        Args:
+            circuitset ([core.circuit.Circuit]): quantum circuits to be executed.
+            qubit_operator(openfermion): Operator for which we calculate the expectation value.
+
+        Returns:
+            [ExpectationValues]: list of objects representing expectation values for given operator.
+        """
+        expectation_values = []
+        for circuit in circuitset:
+            expectation_values.append(self.get_expectation_values(circuit, qubit_operator, **kwargs))
+        return expectation_values
+
+    def get_bitstring_distribution(self, circuit:Circuit, **kwargs) -> BitstringDistribution:
+        """
+        Calculates a bitstring distribution.
+
+        Args:
+            circuit (core.circuit.Circuit): quantum circuit to be executed.
+
+        Returns:
+            BitstringDistribution: object representing the probabilities of getting specific bistrings.
+
+        """
+        # Get the expectation values
+        measurements = self.run_circuit_and_measure(circuit, **kwargs)
+        return measurements.get_distribution()
+
 
 class QuantumSimulator(QuantumBackend):
     @abstractmethod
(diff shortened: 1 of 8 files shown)


src/python/zquantum/core/interfaces/backend.py
```diff
@@ -41,6 +41,38 @@ def get_expectation_values(self, circuit:Circuit, qubit_operator:QubitOperator,
         """
         raise NotImplementedError
 
+    def get_expectation_values_for_circuitset(self, circuitset:[Circuit], qubit_operator:QubitOperator, **kwargs) -> [ExpectationValues]:
```
mstechly, line 44: Actually, according to https://docs.python.org/3/library/typing.html, it should be:
`circuits:List[Circuit]` (with `from typing import List` at the top)

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -0,0 +1,160 @@
...
+from pyquil import Program
+from pyquil.gates import X, CNOT, H
+from pyquil.wavefunction import Wavefunction
+from openfermion import QubitOperator, IsingOperator
+
+from ..circuit import Circuit
+from ..measurement import Measurements, ExpectationValues
+from ..bitstring_distribution import BitstringDistribution
+
+class QuantumBackendTests(object):
```
mstechly, line 12: Why do we inherit after `object` and not `unittest.TestCase`?
Just asking.

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -0,0 +1,160 @@
...
+        # When
+        for backend in self.backends:
+            backend.n_samples = n_samples
+            measurements = backend.run_circuit_and_measure(circuit)
+
+            # Then (since SPAM error could result in unexpected bitstrings, we make sure the most common bitstring is
+            #   the one we expect)
+            counts = measurements.get_counts()
+            self.assertEqual(max(counts, key=counts.get), "001")
+
```
mstechly, line 30: There should be just one line of break between two methods of the same class ;)
https://www.python.org/dev/peps/pep-0008/#blank-lines

mstechly: Elsewhere as well.

src/python/zquantum/core/interfaces/backend.py
simonwa7: Addressed here:
68c0a60

src/python/zquantum/core/interfaces/backend_test.py
simonwa7: This was a mistake on my part. Addressed here:
5abb85a

src/python/zquantum/core/interfaces/backend_test.py
simonwa7: Addressed here:
7dcdc1e

src/python/zquantum/core/interfaces/backend_test.py
simonwa7: Actually, previous commit is reverted in this commit: e39c164. This is because `pytest` will attempt to run these base classes as test cases themselves instead of ignoring them.

simonwa7:
Added some basic test cases here: 8170d63. I think if we want to increase our test coverage, we should leave that for a future story.

src/python/zquantum/core/interfaces/backend_test.py
mstechly: I see 👍

src/python/zquantum/core/interfaces/optimizer_test.py
```diff
@@ -22,3 +25,16 @@ def test_optimization(self):
             self.assertIn("opt_params", results.keys())
             self.assertIn("history", results.keys())
 
+    def test_optimization_simple_function(self):
+        for optimizer in self.optimizers:
+            results = optimizer.minimize(simple, initial_params=[1, -1])
```
mstechly, line 30: I think it should not work :) 
Simple is a python function, not `CostFunction` object and therefore it will fail.
Also, actually `MockCostFunction` implements exactly this.

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -49,28 +49,39 @@ def test_run_circuit_and_measure_correct_num_measurements(self):
     def test_get_expectation_values(self):
         # Given
         circuit = Circuit(Program(H(0), CNOT(0,1), CNOT(1,2)))
-        qubit_operator = IsingOperator('[]')
+        operator = IsingOperator('[]')
```
mstechly, line 52: This is an empty operator and then you have a `test_get_expectation_values_empty_op`, so it's slightly confusing for me.

src/python/zquantum/core/interfaces/optimizer_test.py
simonwa7: Addressed here: 
ec150a4 

Not using the `MockCostFunction` because in theory, we don't know exactly what that cost function is actually doing and here we are making sure it's finding the correct parameters for a known problem.

src/python/zquantum/core/interfaces/backend_test.py
simonwa7: It's actually not an empty operator. It's an operator with only the identity term. I think this test is actually useful to make sure the identity case is handled properly

src/python/zquantum/core/interfaces/optimizer_test.py
mstechly: Makes sense 👍

src/python/zquantum/core/interfaces/backend_test.py
mstechly: 👍

PR #22, opened 2020-05-22
Error wherein parameters used for evaluation of gradient were incorre…

simonwa7:
…ctly assigned


src/python/zquantum/core/cost_function_test.py
```diff
@@ -61,8 +61,8 @@ def test_get_finite_difference_gradient(self):
         function = np.sum
         params_1 = np.array([1,2,3])
         params_2 = np.array([1,2,3,4])
-        target_gradient_value_1 = np.array([1,1,1])
-        target_gradient_value_2 = np.array([1,1,1,1])
+        target_gradient_value_1 = np.zeros(3)
```
mstechly, line 64: I think the previous version of tests was actually correct and the problem lies on the side of the gradient method.
The gradient values should be same as the values for the analytical gradient (up to precision).

PR #35, opened 2020-06-10
Added tests for gate class

mstechly:


src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+        return gate, qubit_list
+
+    def create_gate_with_symbolic_params(self, gate_name):
+        if gate_name in self.two_qubit_gates:
+            qubit_list = [Qubit(0), Qubit(1)]
+        else:
+            qubit_list = [Qubit(0)]
+        params = []
+        if gate_name in self.one_parameter_gates:
+            params = [Symbol("theta_0")]
```
simonwa7, line 35: Do we have any two parameter gates?

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+        self.assertEqual(evaluated_regular_gate, gate)
+
+        # Given
+        symbols_map = [("x", 1.0)]
+
+        # When
+        evaluated_symbolic_gate = symbolic_gate.create_evaluated_gate(symbols_map)
+
+        # Then
+        self.assertEqual(evaluated_symbolic_gate, symbolic_gate)
```
simonwa7, line 63: I feel like there are two different tests here. Maybe split this function up into the "normal" `gate` and `symbolic_gate`

simonwa7: Also maybe loop over gates in `COMMON_GATES`?

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+        if gate_name in self.two_qubit_gates:
+            qubit_list = [Qubit(0), Qubit(1)]
+        else:
+            qubit_list = [Qubit(0)]
+        if params is None:
+            params = []
+            if gate_name in self.one_parameter_gates:
+                params = [1.0]
+        gate = Gate(gate_name, qubits=qubit_list, params=params)
+        return gate, qubit_list
```
simonwa7, line 26: Why do we return `qubit_list`? Shouldn't it be a member of the gate object already?

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+
+    def create_gate_with_symbolic_params(self, gate_name):
```
simonwa7, line 28: Same comments as for `create_gate`

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+            self.assertEqual(gate, recreated_gate)
+
+    def test_pyquil_io(self):
+        for gate_name in COMMON_GATES:
+            # Given
+            gate, qubit_list = self.create_gate(gate_name)
+
+            # When
+            pyquil_gate = gate.to_pyquil()
+            recreated_gate = Gate.from_pyquil(pyquil_gate, qubit_list)
```
simonwa7, line 84: Don't pyquil gates have the qubit list as a member of the object? This seems problematic if not...

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+            self.assertEqual(gate, recreated_gate)
+
+    def test_cirq_io(self):
+        for gate_name in COMMON_GATES:
+            # Given
+            gate, qubit_list = self.create_gate(gate_name)
+
+            # When
+            cirq_gate = gate.to_cirq()
+            recreated_gate = Gate.from_cirq(cirq_gate, qubit_list)
```
simonwa7, line 96: Don't cirq gates have the qubit list as a member of the object? This seems problematic if not...

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+            gate_evaluated = gate.create_evaluated_gate(symbols_map)
+            recreated_gate_evaluated = recreated_gate.create_evaluated_gate(symbols_map)
+
+            # Then
+            # There were numerical & sympy related issues when comparing gates directly, so in this case we compare the evaluated forms of the gates.
+            self.assertEqual(gate_evaluated, recreated_gate_evaluated)
```
simonwa7, line 181: Because of this, I think we should test a couple different assignments for the parameter. Including corner cases such as `0`, `pi`, `-pi`, `2pi`, etc

src/python/zquantum/core/circuit/_gateset.py
```diff
@@ -19,26 +19,4 @@
...
-CIRQ_GATES = ['PhasedXPowGate', 'CNOT', 'CNotPowGate', 'CZ', 'CZPowGate', 'H', 'HPowGate', 
-              'ISWAP', 'ISwapPowGate', 'measure', 'measure_each', 'MeasurementGate', 'XPowGate', 
-              'YPowGate', 'ZPowGate', 'Rx', 'Ry', 'Rz', 'S', 'SWAP', 'SwapPowGate', 'T', 'X', 'Y', 
-              'Z', 'XX', 'XXPowGate', 'YY', 'YYPowGate', 'ZZ', 'ZZPowGate', 'PhasedXPowGate',
-              'CCX', 'CCXPowGate', 'CCZ', 'CCZPowGate', 'CSWAP', 'CSwapGate', 'FREDKIN', 'TOFFOLI']
-                        
-# All gates natively support in qiskit
-QISKIT_GATES = ['ubase', 'u2', 'h', 'cxbase', 'cx', 'u1', 't', 'ccx', 'cswap', 'cx_base', 's', 'cy', 
-                'cz', 'swap', 'iden', 'sdg', 'tdg', 'u0', 'u3', 'u_base', 'x', 'y', 'z', 'rx', 'ry', 
-                'rz', 'cu1', 'ch', 'crz', 'cu3', 'rzz', 'measure', 'barrier']
```
simonwa7, line 44: Do we not use these anywhere else?

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
\ No newline at end of file
+    def setUp(self):
+        self.two_qubit_gates = ["CNOT", "CZ", "CPHASE", "SWAP"]
+        self.one_parameter_gates = ["PHASE", "Rx", "Ry", "Rz", "CPHASE"]
+
+    def create_gate(self, gate_name, params=None):
+        if gate_name in self.two_qubit_gates:
+            qubit_list = [Qubit(0), Qubit(1)]
+        else:
+            qubit_list = [Qubit(0)]
```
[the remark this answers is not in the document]
mstechly: Adressed in `2cb14ce59e860ea43767b0aa336fa0da102b99dc`

src/python/zquantum/core/circuit/_gate_test.py
```diff
@@ -1,7 +1,184 @@
...
+        if params is None:
+            params = []
```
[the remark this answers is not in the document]
mstechly: Because there is this issue when you pass empty lists or dictionaries as input params in Python. I don't remember why exactly it causes problems, but it's better not to do that.

src/python/zquantum/core/circuit/_gate_test.py
mstechly: We do have (e.g.: `ZXZ`), but they're not in `COMMON_GATES` so I left them out for now. But I left the structure, so it's easier to expand the tests when we want to decide to do so.

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `dd8660d6b835c6f15c0e4da35d39e8da14ffbf78`

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `01b1a27f3c57ca60d3707f7a8497e7a50a6322f7`

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `01b1a27f3c57ca60d3707f7a8497e7a50a6322f7`

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `26e22a34f6d398f0145ca301500ecb2964a0dc40`

src/python/zquantum/core/circuit/_gateset.py
mstechly: Anywhere at all. Seems like some legacy code to me.

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `e8e0349b7fc7d913a27ef8e9f2457614f89a8e0b`

src/python/zquantum/core/circuit/_gate_test.py
mstechly: Done in `e8e0349b7fc7d913a27ef8e9f2457614f89a8e0b`

src/python/zquantum/core/circuit/_gate_test.py
simonwa7: Works for me

src/python/zquantum/core/circuit/_gateset.py
simonwa7: Alrighty then - time to get rid of it :D

src/python/zquantum/core/circuit/_gate_test.py
simonwa7: Story added to tracker

src/python/zquantum/core/circuit/_gate_test.py
simonwa7: Story added to tracker

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -84,11 +84,16 @@ def __eq__(self, another_gate):
...
+    def evaluate(self, symbols_map):
         """
-        Creates a copy of gate with symbolic params
+        Returns a copy of a gate with specified symbolic parameters evaluated to provided values.
+
+        Args:
+            symbols_map list(tuple(sympy.Basic, number)): List containing symbols and values that they should take.
         """
-        evaluated_gate = copy.deepcopy(self)
+        gate_class = type(self)
```
simonwa7, line 94: What is this line for?

mstechly: Instead of making a copy (in particular a deep copy, which copies all the referenced objects etc.), we create a new object, which is of the same class as the initial one (and not Gate class, in case you ever have an inherited class).

simonwa7: Yes, but you don't use the `gate_class` variable after you assign it

mstechly: I do:
Screen Shot 2020-06-12 at 10 48 00 AM

PR #37, opened 2020-06-12
Blackify code

simonwa7:


simonwa7:
I haven't tried these specifically, but have used others after being formatted by black

PR #38, opened 2020-06-12
Noise models and device connectivities

simonwa7:


src/python/zquantum/core/interfaces/mock_objects.py
```diff
@@ -64,3 +64,7 @@ def get_gradient(self, parameters):
 
 def mock_ansatz(parameters):
     return Circuit(Program(X(0)))
+
+
+def mock_load_noise_model(noise_model_data):
+    return None
```
mstechly, line 70: This looks like mock which doesn't really helps in testing anything?

mstechly: Also, since it's not an interface, I would move it from `interfaces/mock_objects.py` to `testing`.'

mstechly: Also, the name is confusing!
Cause you have `load_noise_model` in `utils` as "load noise model **from file**".
But this one is "load noise model **from data**".
I'd suggest naming it `create_noise_model_from_data` or something like this?

simonwa7: - It's used here: `test_noise_model_io`
- I initially put it in `mock_objects` because we have `mock_ansatz` there and this is serving a very similar purpose
- I agree about the naming. Will change it to `mock_create_noise_model`

simonwa7: Renamed in this commit: ecaf79183ca23a0ab42635874fa50e971e6ac0a3

mstechly: Regardging the place where it lives – well, we aim for the ansatz to have its own interface, so that's the reason it's in `interfaces/mock_objects.py`
We don't aim for having interface for the noise model (or do we?) so that's why I suggested moving it elsewhere.
But it's a minor thing, do as you wish, approving anyway.

simonwa7: Addressed in commit: 6157b36

PR #39, opened 2020-06-15
172388054 parametric circuits

mstechly:


src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -96,13 +123,44 @@ def test_circuit_init(self):
...
+        target_circuit = Circuit(
+            Program().inst(RX(2 * value_1 + value_2, 0), RY(value_1, 0), RZ(value_2, 0))
+        )
+
+        # When
+        evaluated_circuit = circuit.evaluate(symbols_map)
+
+        # Then
+        self.assertEqual(evaluated_circuit, target_circuit)
+
```
simonwa7, line 145: There should also be test cases for:
- calling `evaluate` on a circuit without any symbolic params
- calling `evaluate` with some but not all of the symbolic params specified
- calling `evaluate` with an empty symbols map or a map with symbols not in the circuit -> what is the intended behavior here
- calling `evaluate` with more symbols in the map than in the circuit

simonwa7:
Additionally, I think it would be good to have a field in the circuit that lists the symbols/parameterized gates in the circuit. I'm sure it will be useful for future development and removes the burden of remembering how many parameterizable gates are in the circuit, which gates have parameters, and which symbols go with which gate

Something like:
```
> print(circuit.parameterized_gates)
{
  'theta_1': [Gate(qubits=(0), parameters=(theta_1), index=10)],
  'theta_2': [Gate(qubits=(0), parameters=(theta_2), index=15), Gate(qubits=(1), parameters=(theta_2), index=16)]
}
```

src/python/zquantum/core/circuit/_circuit_test.py
mstechly: Adressed in [a2bda64](https://github.com/zapatacomputing/z-quantum-core/pull/39/commits/a2bda6408bda9011406fc198d2b194ee04b7d6f8)

mstechly:
@simonwa7 added `symbolic_params` property for gates and circuits in [8cc755](https://github.com/zapatacomputing/z-quantum-core/pull/39/commits/8cc7550c719589998f13f8d5adf672749e591676) and [1620f8](https://github.com/zapatacomputing/z-quantum-core/pull/39/commits/1620f89eb440a74dd952869c3581a4839f6876ce)

src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -96,13 +123,124 @@ def test_circuit_init(self):
...
+        # When
+        symbolic_params_1 = circ_1.symbolic_params
+        symbolic_params_2 = circ_2.symbolic_params
+
+        # Then
+        self.assertEqual(set(symbolic_params_1), set(target_symbolic_params_1))
+        self.assertEqual(symbolic_params_2, target_symbolic_params_2)
+
+    def test_circuit_evaluate_with_all_params_specified(self):
```
simonwa7, line 156: This test is actually not "all_params", but more than all params

src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -96,13 +123,124 @@ def test_circuit_init(self):
...
+        circ_2.qubits = qubits
+        circ_2.gates = [gate_4]
+
+        # When
+        symbolic_params_1 = circ_1.symbolic_params
+        symbolic_params_2 = circ_2.symbolic_params
+
+        # Then
+        self.assertEqual(set(symbolic_params_1), set(target_symbolic_params_1))
+        self.assertEqual(symbolic_params_2, target_symbolic_params_2)
```
simonwa7, line 154: This seems more like two separate unittests -> circuit with no params, circuit with some params. Although I don't think it's necessary to change

src/python/zquantum/core/circuit/_circuit.py
```diff
@@ -81,13 +81,36 @@ def n_multiqubit_gates(self):
...
+    def symbolic_params(self):
+        """
+        Returns a list of symbolic parameters used in the circuit
+        """
+        symbolic_params = []
+        for gate in self.gates:
+            symbolic_params_per_gate = gate.symbolic_params
+            symbolic_params += symbolic_params_per_gate
+
+        return symbolic_params
```
simonwa7, line 94: This doesn't give any information about gate associated with the symbol. I think that information would be useful to have, but open to hearing why that may not make sense

src/python/zquantum/core/circuit/_circuit_test.py
mstechly: Naming is confusing – method is called `symbolic_params` and it's method of `circuit` and it's `test`, so altogether it's `test_circuit_symbolic_params`, but the test actually tests it for circuit with and without symbolic params, to make sure the behaviour is correct if circuit is "regular".

src/python/zquantum/core/circuit/_circuit_test.py
mstechly: Well, but all params of this circuit are specified, aren't they?

src/python/zquantum/core/circuit/_circuit.py
mstechly: If we ever need it, we can add it then.
If we never use it, the code will linger there forever ;) 
And honestly, I don't see a clear use case why I would care to know which gates are with symbolic parameters – in case of complicated ansatzes knowing that there's a parametrizable `RX` doesn't really provide me a lot of information. And which `RX` exactly is that? The first one or the 5th from the end?

src/python/zquantum/core/circuit/_circuit_test.py
simonwa7: True, but it's misleading. I think there are two different test cases here. One where all the parameters in the circuit, but no more, are specified and one where all the parameters, including extra parameters, are specified. I could see this second scenario raising a warning as the intended behavior (possibly in this PR or in the future).

src/python/zquantum/core/circuit/_circuit_test.py
simonwa7: Good point - maybe these should actually be assertions in `test_circuit_init` and `test_circuit_init_with_symbolic_params`?

src/python/zquantum/core/circuit/_circuit.py
simonwa7: Fair point - I guess I could see it being useful, but I can't say for sure.

src/python/zquantum/core/circuit/_circuit_test.py
mstechly: Fixed it in [d86b61](https://github.com/zapatacomputing/z-quantum-core/pull/39/commits/d86b61466a3b925de0fe012d353149b75b0bc251)

src/python/zquantum/core/circuit/_circuit_test.py
mstechly: I've updated the tests and added a warning, see if it makes more sense now.

PR #40, opened 2020-06-15
Adding template to create a random circuit

simonwa7:


templates/circuit.yaml
```diff
@@ -158,3 +158,28 @@ spec:
...
+            default: "None"
+          - name: command
+            value: python3 main_script.py
+        artifacts:
+          - name: main-script
+            path: /app/main_script.py
+            raw:
+              data: |
+                from zquantum.core.circuit import save_circuit
+                from zquantum.core.testing import create_random_circuit
```
mstechly, line 178: Since it seems we'll start using this function outside of testing, perhaps it would make sense to move it from `testing` to `utils`?

simonwa7: Even though this function is now used inside a template, it will be used primarily for testing other templates.

PR #41, opened 2020-06-15
Installing qutip in default docker image

simonwa7:


docker/z-quantum-default/0.0.4/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+                    cffi>=1.14.0 \
+                    scipy==1.4.1 \
+                    lea==3.2.0 \
+                    numpy==1.18.1 \
+                    werkzeug==0.14.1 \
+                    flask==1.1.1 \
+                    pyyaml==5.1 \
+                    quantum-grove>=1.0.0
+
+RUN python3 -m pip install qutip
```
mstechly, line 74: Why as a separate command instead of spending it to the previous one?

simonwa7: There were some issues popping up during the `build` process when it was part of the previous command.

mstechly: And why no version is specified for qutip?

simonwa7: Addressed here: 33c34cf

PR #44, opened 2020-06-17
Introduce estimators and support them in cost functions

sabeegrewal:
In this PR:
* Introduce the `Estimator` interface. 
* Implement `ExactEstimator` and `BasicEstimator`. `ExactEstimator` uses `get_exact_expectation_values` from a backend. `BasicEstimator` uses `get_expectation_values` from a backend. 
* Rename `EvaluateOperatorCostFunction` to `AnsatzBasedCostFunction`.  
* `AnsatzBasedCostFunction` now takes an `Estimator` (and it's `BasicEstimator` by default). 
* `_evaluate` and `evaluate` return `ValueEstimates` in `CostFunction`. 
* Override `__eq__` method in `ValueEstimate`
* Modify existing `CostFunction` tests 
* Switch to `MockEstimator` in `cost_function_tests`
* Add tests for `Estimators`
* Add sample complexity, error, and confidence to the estimator interface

Deferred to a later PR: 
* return precision in `get_estimated_expectation_values`.


src/python/zquantum/core/estimator.py
```diff
@@ -0,0 +1,62 @@
...
+            target_operator (List[SymbolicOperator]): List of target functions to be estimated.
+
+        Returns:
+            ExpectationValues: expectation values for each term in the target operator.
+        """
+        return backend.get_expectation_values(circuit, target_operator)
+
+
+class ExactEstimator(Estimator):
+    """An estimator that must run on a quantum simulator. The estimator exactly computes the expectation values an operator.
```
mstechly, line 33: "values **of** an operator"

mstechly: Actually even "of **the** operator".

src/python/zquantum/core/cost_function.py
```diff
@@ -83,14 +86,15 @@ def get_gradient(self, parameters: np.ndarray) -> np.ndarray:
             raise Exception("Gradient type: %s is not supported", self.gradient_type)
 
 
-class EvaluateOperatorCostFunction(CostFunction):
+class AnsatzBasedCostFunction(CostFunction):
```
mstechly, line 89: A think tests should be updated appropriately.

mstechly: Here and elsewhere as well.

sabeegrewal: Will do. Thanks for taking the first pass, I'll do tests tomorrow 😄

src/python/zquantum/core/cost_function.py
```diff
@@ -134,8 +141,8 @@ def _evaluate(self, parameters: np.ndarray) -> float:
...
         """
         circuit = build_ansatz_circuit(self.ansatz, parameters)
-        expectation_values = self.backend.get_expectation_values(
-            circuit, self.target_operator
+        expectation_values = self.estimator.get_estimated_values(
+            self.backend, circuit, self.target_operator
         )
         final_value = np.sum(expectation_values.values)
-        return final_value
+        return ValueEstimate(final_value)
```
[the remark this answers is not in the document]
max-radin: We could estimate the precision of the value estimate using the covariances in the `ExpectationValues` object returned by the estimator. However I think it's ok to merge this PR without adding that - computing precisions could be a separate story.

src/python/zquantum/core/cost_function.py
```diff
@@ -111,19 +116,21 @@ def __init__(
         target_operator: SymbolicOperator,
         ansatz: Dict,
         backend: QuantumBackend,
+        estimator: Estimator = BasicEstimator(),
```
max-radin, line 119: Generally we don't want default values to be mutable types because it can lead to unexpected behavior. You can instead have the default value be `None` and have an if statement in the initializer:
```python
if estimator is None:
  estimator = BasicEstimator()
```

src/python/zquantum/core/cost_function.py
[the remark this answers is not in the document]
mstechly: Ok, sounds good 👍

mstechly:
Looks good to me, though I wouldn't merge it until we have PRs in `z-quantum-optimizers` and `z-quantum-qcbm` – it sometimes turns out that when you propagate changes, you realize that actually you still need to introduce some changes in the core.

src/python/zquantum/core/estimator.py
```diff
@@ -22,10 +25,25 @@ def get_estimated_expectation_values(
             backend (QuantumBackend): the backend that will be used to run the circuit
             circuit (Circuit): the circuit that prepares the state.
             target_operator (List[SymbolicOperator]): List of target functions to be estimated.
+            n_samples (int): Number of measurements done on the unknown quantum state.
```
mstechly, line 28: I would leave out "on the unknown quantum state", but it's super minor ;)

src/python/zquantum/core/interfaces/estimator.py
```diff
@@ -27,8 +31,33 @@ def get_estimated_expectation_values(
...
+            n_samples (int): Number of measurements done on the unknown quantum state. 
+            epsilon (float): an error term.
+            delta (float): a confidence term.
 
         Returns:
             ExpectationValues: The estimations of the terms in the target_operator. 
         """
         raise NotImplementedError
+
+    def _ignore_parameter(
```
mstechly, line 43: Don't you think something like "_log_ignore_parameter" or "_log_parameter_ignored" is a better name?

src/python/zquantum/core/estimator.py
```diff
@@ -22,10 +25,25 @@ def get_estimated_expectation_values(
...
 
         Returns:
             ExpectationValues: expectation values for each term in the target operator.
         """
+        estimator_name = type(self).__name__
+        self._ignore_parameter(estimator_name, "epsilon", epsilon)
+        self._ignore_parameter(estimator_name, "delta", delta)
+
+        if n_samples is not None:
+            save = backend.n_samples
```
mstechly, line 40: Not a very good variable name – I have no idea what it means.

src/python/zquantum/core/estimator.py
```diff
@@ -22,10 +25,25 @@ def get_estimated_expectation_values(
...
+            backend.n_samples = n_samples
+            expectation_values = backend.get_expectation_values(
```
mstechly, line 42: This whole `if` is confusing to me – why do you calculate expectation values if you don't use them at all?

src/python/zquantum/core/estimator.py
mstechly: I'd throw a warning here (see other comments).

src/python/zquantum/core/interfaces/estimator.py
```diff
@@ -11,14 +12,17 @@ class Estimator(ABC):
...
-        pass
+        self.logger = logging.getLogger("z-quantum-core")
 
     @abstractmethod
     def get_estimated_expectation_values(
         self,
         backend: QuantumBackend,
         circuit: Circuit,
         target_operator: SymbolicOperator,
+        n_samples: int = None,
```
mstechly, line 23: I wonder if we couldn't put that in `__init__` ? Not sure if it makes more sense to have it here or separated...
One thing I don't like is the fact, that now we have n_samples specified in two places – in the definition of backend and here. And it's not clearly indicated that n_samples passed here overrides the n_samples from backend.

I wonder what would be the best way to deal with that... Perhaps we shouldn't pass n_samples and just assume we use the value from backend. Then we have a single source of truth, but I think it might be a little bit cumbersome to change the backend object every time before we want to run get_estimated_expectation_values.
So perhaps it's better to do it this way, just explicitly say in the docs that the n_samples here overrides the one defined in backend and throw a warning if the two don't match?

WDYT @max-radin @simonwa7 ?

src/python/zquantum/core/interfaces/estimator_test.py
```diff
@@ -24,10 +58,45 @@ def test_get_estimated_expectation_values(self):
...
+):
+    with self.assertLogs("z-quantum-core", level="WARN") as context_manager:
+        # Given
+        expected_logs = [
+            "WARNING:z-quantum-core:{} = {}. {} does not use {}. The value was ignored.".format(
+                parameter_name, parameter_value, estimator_name, parameter_name
+            )
+        ]
+        # When
+        if parameter_name == "n_samples":
```
mstechly, line 86: Why do you override the self values of the parameters instead of just passing n_samples=n_samples in the estimator.get_estimated_expectation_values ?

src/python/zquantum/core/estimator.py
sabeegrewal: The line reads as "save backend.n_samples" because it saves backend.n_sample :) it's as clear as you can get 😆  Oh well, I figured someone would leave a comment... I'll make the variable name clearer (rather than the line itself)

sabeegrewal: I'll throw a warning too.

src/python/zquantum/core/estimator.py
sabeegrewal: the return statement should be returning `expectation_values`. just a bug, will fix in the next commit.

src/python/zquantum/core/interfaces/estimator.py
sabeegrewal: `n_samples` is something that can change per estimation routine, which is why I made it a method parameter. 

I don't think `QuantumBackend` should have `n_samples` as an attribute. Whether I'm right or wrong, I have no idea. But, I also think this PR is getting too big so I made the estimator take `n_samples` in a way that causes no side effects.

In the future, I would like to see objects using the backend tell it how many times to run something (rather than there being a single attribute that is set on initialization). But, again, I'm not sure if that's the way to go and left it for a future PR.

src/python/zquantum/core/interfaces/estimator_test.py
sabeegrewal: It depends on the parameter being ignored. When `parameter_name` is `"n_samples"`, then I want to do `n_samples=parameter_value`. Otherwise, I want to do `n_samples=self.n_samples'.

mstechly: Oh, makes sense.

src/python/zquantum/core/interfaces/estimator.py
mstechly: Makes sense 👍

src/python/zquantum/core/estimator.py
mstechly: Typo in `arugment` ;)

PR #45, opened 2020-06-19
Mct

akataba:
Added capability for multi-control gates in Gate Class

### DIFF
--- a/src/python/zquantum/core/circuit/_circuit_test.py
+++ b/src/python/zquantum/core/circuit/_circuit_test.py
@@ -13,7 +13,8 @@ from pyquil.gates import *
 from pyquil.gates import QUANTUM_GATES
 from math import pi, sin, cos
 
-from . import load_circuit, save_circuit, Circuit, Gate, Qubit, pyquil2cirq, cirq2pyquil
+from . import load_circuit, save_circuit, Circuit, Gate, Qubit, pyquil2cirq, cirq2pyquil,\
+        MCTGate, PhaseOracle, MCRY
 
 from ..utils import compare_unitary, is_identity, is_unitary, RNDSEED
 from ..testing import create_random_circuit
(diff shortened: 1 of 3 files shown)


src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +915,116 @@ def _unitary_(self):
...
+            self.ccx_list = []
+            self.qiskit_data = []
+            for i in range(len(self.all_qubits)):
+                if i not in self.control_qubits and i not in self.target_qubits:
+                    self.ancilla_qubits.append(i)
+        else:
+            raise TypeError('Expecting a list integer labels for control and target qubits')
+        # This is the ibm circuit object with all the toffoli gates
+        self.ccx_decomposition = qiskit.QuantumCircuit(len(self.all_qubits), len(self.all_qubits))
+        self.qiskit_ctrl_q = None
```
mstechly, line 967: Shouldn't these be actually private?

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +915,116 @@ def _unitary_(self):
...
+
+    qiskit_ctrl_q (list): List of qiskit qubits that are control qubits for the MCT gate
+    qiskit_targ_q (list): List of qiskit qubits that are targ qubits for the MCT gate
+
+    qiskit_ancilla_q (list): List of qiskit qubits that are ancilla qubits for the mct gate synthesis
+
+
+    """
+
+    def __init__(self, circuit, control_qubits=[], target_qubits=[]):
```
mstechly, line 951: You should not use empty lists as default parameters – it might lead to weird behaviour and very hard to track bugs. It is much better to do:
```
def __init__(self, control_qubits=None):
    if control_qubits is None:
        control_qubits = []
```

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +915,116 @@ def _unitary_(self):
...
+            self.ancilla_qubits = []
+            self.ccx_list = []
+            self.qiskit_data = []
+            for i in range(len(self.all_qubits)):
+                if i not in self.control_qubits and i not in self.target_qubits:
+                    self.ancilla_qubits.append(i)
+        else:
+            raise TypeError('Expecting a list integer labels for control and target qubits')
+        # This is the ibm circuit object with all the toffoli gates
+        self.ccx_decomposition = qiskit.QuantumCircuit(len(self.all_qubits), len(self.all_qubits))
```
mstechly, line 966: Given how you create it now, `self.ccx_decomposition` is an empty circuit which doesn't hold any value.
Therefore, it doesn't really make sense to initialize it now – it would make more sense to set it to `None` and initialize it in `_synthesize_circuit`, when you actually need to use it. Right?

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +915,116 @@ def _unitary_(self):
...
+    def _synthesize_circuit(self):
+        """
+         Produces the gate decomposition using qiskit
+        """
+        self.ccx_decomposition.mct(self.qiskit_ctrl_q, self.qiskit_targ_q, self.qiskit_ancilla_q)
+    def _create_qiskit_qubits(self):
+        """
+        Producing list of qiskit qubit objects
+        """
+        qreg = QuantumRegister(len(self.circuit.qubits), 'q')
```
mstechly, line 981: It seems to me that:
a) You could have used `self.all_qubits` here instead of `self.circuit.qubits`
b) You need to have `circuit` only to get the qubits – so actually, you don't need to pass `circuit` as a parameter (which I really don't like), you just need to pass `qubits`, right?

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +915,116 @@ def _unitary_(self):
...
+        self.qiskit_ancilla_q = [qreg[i] for i in self.ancilla_qubits]
+
+    def create_gate_list(self):
+        # First create qiskit objects
+        self._create_qiskit_qubits()
+
+        # Synthesize circuit
+        self._synthesize_circuit()
+        for i in range(len(self.ccx_decomposition.data)):
+            self.qiskit_data += list(self.ccx_decomposition[i])
```
mstechly, line 993: It doesn't return anything, is that intended?
If so, it's confusing.

Also, this method doesn't really take any arguments, so it seems that you have all the information to execute it when you run init – why don't make it private and just run it on init?

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -616,6 +620,21 @@ def to_qiskit(self, qreg=None):
...
+        if self.name == 'CCX':
+            return [qiskit.extensions.standard.ToffoliGate(), 
+            [qiskit_qubits[0], qiskit_qubits[1], qiskit_qubits[2]], []]
+        if self.name == 'MCT':
+            gate = MCTGate(self.circuit, control_qubits=self.control_qubits, target_qubits=self.target_qubits)
+            gate.create_gate_list()
+            return gate.qiskit_data
+        if self.name == 'MCU1':
+            gate = PhaseOracle(self.params[0], self.circuit, control_qubits=self.control_qubits, target_qubits=self.target_qubits)
+            gate.create_gate_list()
```
mstechly, line 632: I think `PhaseOracle` doesn't have `create_gate_list` method?

src/python/zquantum/core/circuit/_gate.py
akataba: I don't have all the information to run it when I run init, I need to create qiskit qubits first.

src/python/zquantum/core/circuit/_gate.py
akataba: If I do that I will have a variable that starts of as None. But then later I will then need to put in an empty circuit, then put in qiskit qubits. The first step is in the end not needed.

src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -1439,6 +1442,114 @@ def test_exp_S_inverse(self):
...
+        circ.qubits = qubits
+        circ.gates = [CCX]
+        # Change to qiskit version
+        ibm_circ = circ.to_qiskit()
+
+        # Create the same qiskit circuit
+        q_reg = qiskit.QuantumRegister(6, 'q')
+        c_reg = qiskit.ClassicalRegister(6, 'c')
+        circuit = qiskit.QuantumCircuit(q_reg, c_reg)
+        circuit.ccx(2,3,4)
```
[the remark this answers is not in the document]
akataba: Changed variable names to more meaningful ones

src/python/zquantum/core/circuit/_gate.py
mstechly: Ok, makes sense 👍

src/python/zquantum/core/circuit/_gate.py
akataba: I have now called the `create_gate_list` in the `__init__ ` method

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -888,3 +912,127 @@ def _unitary_(self):
...
+    def _create_qiskit_qubits(self):
+        """
+        Producing list of qiskit qubit objects
+        """
+        qreg = QuantumRegister(len(self.all_circuit_qubits), 'q')
+        self.qiskit_ctrl_q = [qreg[i] for i in self.control_qubits]
+        self.qiskit_targ_q = qreg[self.target_qubits[0]]
+        self.qiskit_ancilla_q = [qreg[i] for i in self.ancilla_qubits]
+
+    def create_gate_list(self):
```
mstechly, line 988: I think it should be `_create_gate_list` now?

akataba: changed

PR #47, opened 2020-06-23
Context selection

max-radin:
Updates the `BasicEstimator` to handle non-diagonal operators by applying context selection gates. Each term is computed in a separate circuit; grouping will be implemented in future PRs.

### DIFF
--- a/src/python/zquantum/core/estimator.py
+++ b/src/python/zquantum/core/estimator.py
@@ -2,8 +2,38 @@ from .interfaces.estimator import Estimator
 from .interfaces.backend import QuantumBackend, QuantumSimulator
 from .circuit import Circuit
 from .measurement import ExpectationValues
-from openfermion import SymbolicOperator
+from openfermion import SymbolicOperator, QubitOperator, IsingOperator
 from overrides import overrides
+import numpy as np
+import pyquil
+from typing import Tuple
+
+
+def get_context_selection_circuit(
+    term: Tuple[Tuple[int, str], ...]
+) -> Tuple[Circuit, IsingOperator]:
+    """Get the context selection circuit for measuring the expectation value
+    of a Pauli term.
+
+    Args:
+        term: The Pauli term, expressed using the OpenFermion convention.
+    
+    Returns:
+        Tuple containing:
+        - The context selection circuit.
+        - The frame operator
+    """
+
+    context_selection_circuit = Circuit()
+    operator = IsingOperator(())
+    for factor in term:
+        if factor[1] == "X":
+            context_selection_circuit += Circuit(pyquil.gates.RY(-np.pi / 2, factor[0]))
+        elif factor[1] == "Y":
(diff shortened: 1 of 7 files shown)


src/python/zquantum/core/estimator.py
```diff
@@ -1,9 +1,43 @@
...
+)
+from openfermion import SymbolicOperator, QubitOperator, IsingOperator
 from overrides import overrides
+import numpy as np
+import pyquil
+from typing import Tuple
+
+
+def get_context_selection_circuit(
+    term: Tuple[Tuple[int, str], ...]
```
mstechly, line 17: I think ... is unnecessary, though I don't mind it.

src/python/zquantum/core/estimator.py
```diff
@@ -38,6 +72,13 @@ def get_estimated_expectation_values(
         self._log_ignore_parameter(estimator_name, "epsilon", epsilon)
         self._log_ignore_parameter(estimator_name, "delta", delta)
 
+        frame_operators = []
```
mstechly, line 75: So this is "Basic Estimator using basic grouping technique" right?
Do we plan to use different estimators for doing different grouping approaches?

src/python/zquantum/core/estimator_test.py
```diff
@@ -14,7 +43,7 @@ def setUp(self):
         self.operator = QubitOperator("Z0")
         self.circuit = Circuit(Program(X(0)))
         self.backend = MockQuantumBackend()
-        self.n_samples = None
+        self.n_samples = 10
```
mstechly, line 46: I think we should have also a separate test for testing the case with `self.n_samples=None`?

src/python/zquantum/core/estimator.py
max-radin: Without the `...`, I think that would indicate a tuple of length 1: see https://stackoverflow.com/a/54747327/2687426. Though maybe it should really be `Iterable` instead since it is more general?

src/python/zquantum/core/estimator.py
max-radin: It's measuring each term independently (i.e. no grouping), although in zmachine we called this the "individual" grouping strategy.

I am thinking that grouping (and other Hamiltonian decompositions such as orbital frames) should happen outside of the estimator classes because some decompositions that can be used with more than one estimator. Also because we've found that it can be useful to have several tasks to manipulate and analyze a decomposition before running an optimization.

src/python/zquantum/core/estimator_test.py
max-radin: Added a test in 4bcf9ed5c4e6c77779b66b2964a067e80643f8c0.

I wonder if we should remove `n_samples` as an attribute of backend so that we don't have two separate places to specify the same thing. (Although I don't think we need to resolve that in this PR.)

PR #50, opened 2020-06-25
Add estimation error terms to cost_function

sabeegrewal:
This PR adds estimation terms to the cost_function.

### DIFF
--- a/src/python/zquantum/core/cost_function.py
+++ b/src/python/zquantum/core/cost_function.py
@@ -19,7 +19,7 @@ class BasicCostFunction(CostFunction):
         gradient_function (Callable): function used to calculate gradients. Optional.
         gradient_type (str): parameter indicating which type of gradient should be used.
         save_evaluation_history (bool): flag indicating whether we want to store the history of all the evaluations.
-        accuracy(float): accuracy term used in finite difference approximation, as accuracy tends to 0, the approximation improves. 
+        finite_diff_step_size(float): the step size used in finite difference approximation.
 
     Params:
         function (Callable): see Args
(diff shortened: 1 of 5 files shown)


src/python/zquantum/core/cost_function.py
```diff
@@ -97,7 +93,11 @@ class AnsatzBasedCostFunction(CostFunction):
         estimator: (zquantum.core.interfaces.estimator.Estimator) = estimator used to compute expectation value of target operator 
         gradient_type (str): parameter indicating which type of gradient should be used.
         save_evaluation_history (bool): flag indicating whether we want to store the history of all the evaluations.
-        epsilon (float): epsilon used for calculating gradient using finite difference method.
+        finite_diff_step_size(float): the step size used in finite difference approximation.
+        n_samples (int): number of samples (i.e. measurements) to be used in the estimator. 
+        epsilon (float): an additive/multiplicative error term. The cost function should be computed to within this error term.
```
mstechly, line 98: Shouldn't these be inputs to the Estimator, not Cost Function?

sabeegrewal: They are inputs to the `get_estimated_expectation_values` method in the `Estimator`. The cost function ultimately passes them to the `Estimator`. 

In the context of an optimization loop, the optimizer may adjust error terms (e.g., `epsilon`) after some iterations of the loop and pass the updated error terms to the cost function. That's why the cost function gets these parameters as well.

mstechly: Makes sense – my gut feeling says there's a better way to do that, but I don't have any better alternative, so let it be ;)

PR #53, opened 2020-07-02
MockQuantumSimulator produces more realistic expectation values 

jromerofontalvo:
Adds an estimator of class NoisyEstimator which produces noisy expectation values according to specified precision (epsilon) in the total expected value of the target operator or according to a total budget of measurements (n_samples).

*Update:* NoisyEstimator code has been moved to `zquantum.estimators`, changes in this PR are mainly in MockEstimator to support noisyEstimator tests and generate more realistic expectation values.


src/python/zquantum/core/interfaces/mock_objects.py
```diff
@@ -56,7 +56,23 @@ def run_circuit_and_measure(self, circuit, **kwargs):
 
     def get_expectation_values(self, circuit, operator, **kwargs):
         n_qubits = len(circuit.qubits)
-        values = [random.random() for i in range(n_qubits)]
+        try:
```
mstechly, line 59: It shouldn't be done by "try-except", cause it might lead to erratic behaviour.
It's much better to do it using "if-else" statement and e.g. check for the existence of attribute terms inside the "if".

PR #54, opened 2020-07-06
Making create_layer_of_gates handle parameterized gates

simonwa7:


src/python/zquantum/core/circuit/_circuit_template.py
```diff
@@ -487,13 +487,31 @@ def _build_circuit_layers_and_connectivity_nearest_neighbors(n_qubits):
     return CircuitConnectivity(connectivity), CircuitLayers([even_layer, odd_layer])
 
 
-def create_layer_of_gates(number_of_qubits: int, gate_name: str) -> Circuit:
-    """
-    Creates a circuit consisting of a single layer of specific gate.
+def create_layer_of_gates(
+    number_of_qubits: int, gate_name: str, parameters: np.ndarray = None
```
mstechly, line 491: Please add:
`from typing import Optional` (if not there yet)
and change `parameters: Optional[np.ndarray]=None`

simonwa7: Addressed here: 
81e86e0

PR #55, opened 2020-07-06
Refactor/bitstring tests

dexter2206:
This is a refactor of tests in `bitstring_distribution_test` module.


src/python/zquantum/core/bitstring_distribution_test.py
```diff
@@ -18,261 +26,271 @@
...
+        {"000": 1e-3, "111": 0},
+    ],
+)
+def test_bitstring_distribution_gets_normalized_by_default(distribution):
+    """Constructing bitstring distribution normalizes it by default."""
+    distribution = BitstringDistribution(distribution)
+    assert is_normalized(distribution.distribution_dict)
+
+
+def test_bitstring_distribution_keeps_original_dict_if_normalization_is_not_requested():
```
mstechly, line 279: I would rename "is_not_requested" to "should_not_be_performed".
In the previous test case it's also not "requested", it's happening by default.

dexter2206: I agree this name is better. Fixed in 9212ee0

PR #57, opened 2020-07-09
Dev graphs

FredericSauv:
Add more type of graphs which could be generated (regular and complete).
Add the capability to generate graphs from specs (a dictionary)


src/python/zquantum/core/graph.py
```diff
@@ -97,11 +95,103 @@ def generate_random_graph_erdos_renyi(
...
+        if uniform(0,1) < probability: # with the given probability
             if random_weights:
                 weight = uniform(0, 1)
             else:
                 weight = 1.0
             output_graph.add_edge(pair[0], pair[1], weight=weight)
+    
+    return output_graph
 
+def generate_random_graph_regular(num_nodes: int,
```
mstechly, line 107: I think `generate_random_regular_graph` is actually a better name ;)

src/python/zquantum/core/graph.py
```diff
@@ -97,11 +95,103 @@ def generate_random_graph_erdos_renyi(
...
                 weight = uniform(0, 1)
             else:
                 weight = 1.0
             output_graph.add_edge(pair[0], pair[1], weight=weight)
+    
+    return output_graph
 
+def generate_random_graph_regular(num_nodes: int, 
+                                    degree: int, 
+                                    random_weights: bool = False) -> nx.Graph:
```
mstechly, line 109: It might be worthwhile adding a `seed` parameter right away as well.

src/python/zquantum/core/graph.py
```diff
@@ -97,11 +95,103 @@ def generate_random_graph_erdos_renyi(
...
+    
+    Returns:
+        A networkx.Graph object
+    """
+    output_graph = nx.random_regular_graph(d = degree, n = num_nodes)
+    output_graph = weight_graph_edges(output_graph, random_weights)
+
+    return output_graph
+
+def generate_complete_graph(num_nodes: int,
```
mstechly, line 130: Actually, we don't need this, because you can create Erdos-Renyi graph with edge probability equal to 1 and you'll get the same result.

src/python/zquantum/core/graph.py
```diff
@@ -97,11 +95,103 @@ def generate_random_graph_erdos_renyi(
...
+        probability = graph_specs['probability']
+        graph = generate_random_graph_erdos_renyi(num_nodes, probability, random_weights)
+    
+    elif type_graph == 'regular':
+        degree = graph_specs['degree']
+        graph = generate_random_graph_regular(num_nodes, degree, random_weights)
+    
+    elif type_graph == 'complete':
+        graph = generate_complete_graph(num_nodes, random_weights)
+
```
mstechly, line 196: There should be also `else` statement at the end which should look like this:
```
else:
    raise(NotImplementedError("This type of graph is not supported: ", type_graph)
```

PR #63, opened 2020-07-20
Mmd cost function

martamau:
This PR adds the MMD distance measure. It is in a separate file due to license issues, as discussed already. 
Some changes have been done to the previous distance measure (clipped log likelihood) for consistence.


src/python/zquantum/core/distance_measures.py
```diff
@@ -0,0 +1,108 @@
...
+    for bitstring in all_keys:
+        # Add 0 to the values list whenever a bistrings isn't found among the keys.
+        target_values.append(target_distribution.distribution_dict.get(bitstring, 0))
+        measured_values.append(
+            measured_distribution.distribution_dict.get(bitstring, 0)
+        )
+
+    basis = np.asarray([int(item, 2) for item in all_keys])  # bitstring to int
+    if not hasattr(sigma, "__len__"):
+        K = compute_rbf_kernel(basis, basis, sigma)
```
mstechly, line 103: Can we rename it from `K` to something more meaningful?
I understand why it is called `K` in the other functions and why don't want to change it (though still think it's a bad variable name), but in our code the name should be more descriptive.

src/python/zquantum/core/distance_measures.py
```diff
@@ -0,0 +1,108 @@
...
+):
+    """ Compute the squared Maximum Mean Discrepancy (MMD) distance measure between between a target bitstring distribution
+    and a measured bitstring distribution.
+    Reference: arXiv.1804.04168.
+
+        Args:
+            target_distribution (BitstringDistribution): The target bitstring probability distribution.
+            measured_distribution (BitstringDistribution): The measured bitstring probability distribution.
+            distance_measure_parameters (dict): dictionary containing the relevant parameters for the maximum mean discrepancy, i.e.:
+                                                - sigma: the bandwidth parameter used to compute the gaussian kernel.
```
mstechly, line 82: I would describe `distance_measure_parameters` as a separate section from `Args`, to make it easier to read (especially in cases with more parameters).
It also lacks information about default value.
sth like:
```
distance_measure_parameters:
    sigma (float): the bandwidth parameter used to compute the gaussian kernel. The default value is 1.0 .
```

src/python/zquantum/core/distance_measures.py
```diff
@@ -0,0 +1,108 @@
...
+        gamma = 1.0 / (2 * sigma)
+    except ZeroDivisionError as error:
+        print("Handling run-time error:", error)
+        raise
+    K = np.exp(-gamma * exponent)
+    return K
+
+
+def compute_multi_rbf_kernel(x_i, y_j, sigmas):
+    """ Compute the multi-gaussian (RBF) kernel matrix K, with K_ij = 1/c Sum_n[exp(-gamma_n |x_i - y_j|^2)] and gamma = 1/(2*sigma).
```
mstechly, line 49: I don't see `c` anywhere in the code 🤔

src/python/zquantum/core/distance_measures.py
```diff
@@ -0,0 +1,108 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import numpy as np
+
+
+def compute_rbf_kernel(x_i, y_j, sigma):
```
mstechly, line 27: These could have typing, though since it's external code I understand if you don't want to modify it.

src/python/zquantum/core/bitstring_distribution_test.py
```diff
@@ -153,10 +172,40 @@ def test_uses_epsilon_instead_of_zero_in_target_distribution():
...
+        )
 
-        compute_clipped_negative_log_likelihood(target_distr, measured_dist, epsilon=0.01)
+        log_spy.assert_has_calls(
+            [mock.call(0.1), mock.call(0.9), mock.call(0.01)], any_order=True
+        )
+
+
+@pytest.mark.parametrize(
+    "measured_dist,distance_measure_params,expected_mmd",
```
mstechly, line 186: I don't see a test testing if `sigma` defaults to `1.0` as it is.
I think there should either be a test for that or the default value should be removed.

src/python/zquantum/core/bitstring_distribution.py
```diff
@@ -208,20 +209,22 @@ def get_qubits_number(self):
 
 
 def compute_clipped_negative_log_likelihood(
-    target_distribution, measured_distribution, epsilon=1e-9
+    target_distribution, measured_distribution, distance_measure_parameters
```
mstechly, line 212: I see we don't have typing in this file, but I think this is good time to actually add it :)

src/python/zquantum/core/bitstring_distribution.py
```diff
@@ -208,20 +209,22 @@ def get_qubits_number(self):
 
 
 def compute_clipped_negative_log_likelihood(
```
mstechly, line 211: I'm wondering if we shouldn't move this to a separate directory?
We need to have MMD in a separate file which is called `distance_measures.py` but then this function is still in `bitstring_distribution.py`.
Perhaps it would make sense to actually put it all into `bistring_distribution` directory and the change into separate files?
WDYT @simonwa7 ?

src/python/zquantum/core/distance_measures.py
martamau: Changed it to `kernel_matrix` everywhere (see [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/67915991fc62d165cf76ffa8b80fd848fc9f803d)). I don't think it's a problem to change it since we already made some modifications to the original code!

src/python/zquantum/core/distance_measures.py
martamau: Changed it [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/6f169a7f4ff15f0441b6de762463849ed13bf2ad)

src/python/zquantum/core/distance_measures.py
martamau: `c` is the the normalization factor `len(sigmas)`.
The index `n` runs from 1 to `c`. 
Shall we clarify it in the docstring?

src/python/zquantum/core/distance_measures.py
martamau: I actually added it and then wasn't sure if it was reasonable, but it's probably worth it since you raised it too. Is there any good reasons why it's better not to change the external code, as far as you know?

src/python/zquantum/core/bitstring_distribution_test.py
martamau: I agree. I don't know how to make it here though, since `sigma` is a local variable within the `compute_mmd` function (same for `epsilon` in `compute_clipped_negative_log_likelihood`). Any suggestions? 😅

src/python/zquantum/core/bitstring_distribution.py
martamau: Added in [this commit](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/da4ff1b90615db7d3b26384a1a8dd6e2d09d8903)

src/python/zquantum/core/distance_measures.py
martamau: Updated also for the clipped log likelihood cost function [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/a8f0187dd7936ac3611261cbb14b0adfeba0a1a9)

src/python/zquantum/core/distance_measures.py
mstechly: I see. Yes, I think it should be clarified in the docs :)

src/python/zquantum/core/distance_measures.py
mstechly: One reason I can think of is that it might be harder to update it if there are any changes to the original code that we want to introduce or if we want to compare it with the original code.
Another is that if there's a lot of external code, you might accidentally introduce some errors.
But in this case I think it should be ok.

src/python/zquantum/core/bitstring_distribution.py
mstechly: Thank you ❤️

src/python/zquantum/core/bitstring_distribution_test.py
mstechly: I think you can calculate it twice - once with `sigma` explicitly set to `1.0` and the other time without passing any arguments. 
Then you can compare if the results from both cases match each other.

src/python/zquantum/core/distance_measures.py
martamau: Addressed [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/1c8260e429fa744045ec72ed5aef29dd3f4ab3ff)

src/python/zquantum/core/bitstring_distribution_test.py
martamau: Done it [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/1c3e5168aefda3cdee23839089c41addcfd83c93)

src/python/zquantum/core/distance_measures.py
martamau: Agree. Done it [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/1c3e5168aefda3cdee23839089c41addcfd83c93)

src/python/zquantum/core/bitstring_distribution.py
martamau: I've just seen that in other parts of Orquestra the i/o methods use the `TextIO` type (see [here](https://github.com/zapatacomputing/z-quantum-core/blob/a5f726ca0d2d90145f2300240d76d4c0f0f6be94/src/python/zquantum/core/graph.py#L22)). I've marked it as string and wanted to check what is the best way to go here @mstechly

src/python/zquantum/core/bitstring_distribution_test.py
martamau: Addressed [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/3955916b079b25bef7f8294b03752cb70ae21852)

src/python/zquantum/core/bitstring_distribution.py
simonwa7: Agreed - I think `bitstring_distribution` should be a directory with:
- `zquantum/core/bitstring_distribution/__init__.py` 
- `zquantum/core/bitstring_distribution/bitstring_distribution.py` (or a similar name to house the `BitstringDistribution` class

Then there should be a `distance_measures` subdirectory with:
- `zquantum/core/bitstring_distribution/distance_measures/clipped_negative_log_likelihood.py`
- `zquantum/core/bitstring_distribution/distance_measures/mmd.py`
- `zquantum/core/bitstring_distribution/distance_measures/utils.py`.

martamau: So I tried to refactor (see [here](https://github.com/zapatacomputing/z-quantum-core/pull/63/commits/b220e44fd95457990a4d387b396cbe7e1dce6d43))
I don't see why there should be the `zquantum/core/bitstring_distribution/distance_measures/utils.py` file, as I think all the utils are related to the `BitstringDistribution` class and I would leave them there, but happy to hear what you think about this @simonwa7 @mstechly

simonwa7: This looks good to me - agreed on placement of `utils`. @mstechly if you're good with this then I think it's good to merge!

PR #64, opened 2020-07-24
Make logger in Estimator at the module level.

sabeegrewal:
These changes make the logger at the module level, which is the Pythonic way of doing logging.
We also make the logging helper function a static method since it does not depend on instance variables.
(There is a single unrelated change: we add `EnforceOverrides` to the `Estimator` abstract base class).

### DIFF
--- a/src/python/zquantum/core/interfaces/estimator.py
+++ b/src/python/zquantum/core/interfaces/estimator.py
@@ -4,15 +4,18 @@ from .backend import QuantumBackend
 from ..measurement import ExpectationValues
 from abc import ABC, abstractmethod
 from openfermion import SymbolicOperator
-from typing import List, Optional
+from typing import Optional
+from overrides import EnforceOverrides
+
+logger = logging.getLogger(__name__)
 
 
-class Estimator(ABC):
+class Estimator(ABC, EnforceOverrides):
     """Interface for implementing different estimators.
     """
 
     def __init__(self):
-        self.logger = logging.getLogger("z-quantum-core")
+        pass
 
     @abstractmethod
     def get_estimated_expectation_values(
@@ -40,8 +43,9 @@ class Estimator(ABC):
         """
         raise NotImplementedError
 
+    @staticmethod
     def _log_ignore_parameter(
-        self, estimator_name: str, parameter_name: str, parameter_value: float
+        estimator_name: str, parameter_name: str, parameter_value: float
     ):
         """In an estimator, users can specify the number of samples, an error term, and/or a confidence term. 
         Usually, selecting two terms fixes the third (e.g., chooseing a number of samples and an error term will fix the probability 
@@ -56,7 +60,7 @@ class Estimator(ABC):
             parameter_value (float): The parameter value.
         """
         if parameter_value is not None:
-            self.logger.warning(
+            logger.warning(
                 "{} = {}. {} does not use {}. The value was ignored.".format(
                     parameter_name, parameter_value, estimator_name, parameter_name,
                 )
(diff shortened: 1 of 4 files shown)


mstechly:
Looks good to me, though I have very little experience with logging in Python, so would rather have @dexter2206 take a look.

src/python/zquantum/core/interfaces/estimator.py
```diff
@@ -4,15 +4,18 @@
...
 
-class Estimator(ABC):
+
+class Estimator(ABC, EnforceOverrides):
     """Interface for implementing different estimators.
     """
 
     def __init__(self):
-        self.logger = logging.getLogger("z-quantum-core")
+        pass
```
dexter2206, line 18: We should probably delete `__init__` now that it is empty.

src/python/zquantum/core/interfaces/estimator.py
```diff
@@ -4,15 +4,18 @@
...
 from openfermion import SymbolicOperator
-from typing import List, Optional
+from typing import Optional
+from overrides import EnforceOverrides
 
+logger = logging.getLogger(__name__)
 
-class Estimator(ABC):
+
+class Estimator(ABC, EnforceOverrides):
```
dexter2206, line 13: I might be missing something, but can you clarify how is adding `EnforceOverrides` related to the changes to the logger placement proposed here? Or is it just an unrelated change?

src/python/zquantum/core/interfaces/estimator.py
```diff
@@ -40,8 +43,9 @@ def get_estimated_expectation_values(
         """
         raise NotImplementedError
 
+    @staticmethod
```
dexter2206, line 46: We changed it to a staticmethod because it does not use `self` now, which is good (I would rather extract it to a module-level function, but that's matter of taste). However, when you see how this is called, the `estimator_name` is always set to `type(self).__name__`. What do you think about making it an instance method again but instead of passing `estimator_name`, just use `self` to obtain the name?

src/python/zquantum/core/interfaces/estimator.py
sabeegrewal: Not missing anything! It's an unrelated change. I updated the PR body.

sabeegrewal:
Regarding pythonic way of logging: it's strongly recommended in python docs. See "A good convention to use when naming loggers..." [here](https://docs.python.org/3/howto/logging.html).

src/python/zquantum/core/interfaces/estimator.py
sabeegrewal: Thank you for the comments! I agree with your points :)  

What do you think about just getting rid of this function? I kind of regret adding it and think its sort of overkill as an `Estimator`-specific helper. Maybe in the future, we can have a `logging_util` module which checks if something is `None` and logs it. Something like:

```
# logging_util.py
def ignore_if_not_none(logger, parameter_name, parameter_value)
```

 If you think it makes sense to have, then I'd be happy to move this to a module-level function before this is merged.

dexter2206:
> Regarding pythonic way of logging: it's strongly recommended in python docs. See "A good convention to use when naming loggers..." [here](https://docs.python.org/3/howto/logging.html).


From what I understand, the docs only talk about the naming of the logger (i.e. the argument passed to `getLogger`), whereas what I was curious about is the placement of the logger variable itself at the module level. From my experience there are two opposite conventions:
- Defining logger at the module level (what we do here)
- Obtaining logger when it is needed (e.g. in a function that uses it).

The interesting part is that both of these approaches are almost equal in performance because loggers are cached and calling `getLogger` with the same name returns precisely the same object.

src/python/zquantum/core/interfaces/estimator.py
dexter2206: I agree! So let's just delete it, and we are good to go! :fire:

PR #65, opened 2020-07-26
Add the optimization_result function.

dexter2206:
This adds the `optimization_result` convenience function that should serve as an additional, static safety layer for constructing instances of `scipy.optimize.OptimizeResult` while ensuring it has the required fields set. 

See accompanying tests for usage examples.

Several notes:
- I thought about calling this `create_optimization_result` but decided to go for a shorter name for the following reasons:
  - `create` is kind of an empty word (like get or set)
  - it is more in line with names used through scientific libraries (e.g. you have `numpy.array` not `numpy.create_array`)
  
  Please feel free to suggest a better name if you have one :)
- Tests don't check if you have to pass optimized value and params as keyword arguments, because it is already checked by the static type checker. Test in your IDE what happens if you skip those or pass them as positional arguments.

If this approach gets approved, I will follow up with PR to z-quantum-optimizers so that we actually use the function proposed here.

### DIFF
--- a/src/python/zquantum/core/interfaces/mock_objects.py
+++ b/src/python/zquantum/core/interfaces/mock_objects.py
@@ -1,7 +1,7 @@
 from .ansatz import Ansatz
 from .ansatz_utils import ansatz_property
 from .backend import QuantumBackend, QuantumSimulator
-from .optimizer import Optimizer
+from .optimizer import Optimizer, optimization_result
 from .cost_function import CostFunction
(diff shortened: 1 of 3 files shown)


dexter2206:
> Looks good, but I think you should change the code in the `mock_optimizer` as well :)

Nice catch, I still seem to forget about `mock_objects` module! This should be fixed in 38ee9e5.

PR #67, opened 2020-07-28
Adding option to name a list-type artifact

pediejo:
Added optional functionality to insert a name for a list artifact when saving.

### DIFF
--- a/src/python/zquantum/core/utils.py
+++ b/src/python/zquantum/core/utils.py
@@ -360,7 +360,7 @@ def save_list(array, filename, artifact_name=""):
         artifact_name (str): optional argument to specify the schema name
     """
     dictionary = {}
-    dictionary["schema"] = SCHEMA_VERSION + artifact_name + "-list"
+    dictionary["schema"] = SCHEMA_VERSION + "-" + artifact_name + "-list"
     dictionary["list"] = array
 
     with open(filename, "w") as f:
--- a/src/python/zquantum/core/utils.py
+++ b/src/python/zquantum/core/utils.py
@@ -351,15 +351,16 @@ def load_list(file):
     return data["list"]
 
 
-def save_list(array, filename):
+def save_list(array, filename, artifact_name=""):
     """Save expectation values to a file.
 
     Args:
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/utils.py
```diff
@@ -351,15 +351,16 @@ def load_list(file):
...
     """Save expectation values to a file.
 
     Args:
         array (list): the list to be saved
         file (str or file-like object): the name of the file, or a file-like object
+        artifact_name (str): optional argument to specify the schema name
     """
     dictionary = {}
-    dictionary["schema"] = SCHEMA_VERSION + "-list"
+    dictionary["schema"] = SCHEMA_VERSION + artifact_name + "-list"
```
mstechly, line 363: I think you want to do have "-" in the schema name.

Consider the following syntax:
`"-".join([SCHEMA_VERSION, artifact_name, "list"])

src/python/zquantum/core/utils_test.py
```diff
@@ -155,6 +155,16 @@ def test_list_io(self):
...
         os.remove("list.json")
 
+    def test_named_list_io(self):
+        # Given
+        initial_list = [0.1, 0.3, -0.3]
+        # When
+        save_list(initial_list, "list.json", "number")
+        loaded_list = load_list("list.json")
+        # Then
+        self.assertListEqual(initial_list, loaded_list)
```
mstechly, line 165: You're not checking if the schema value is properly saved here :)

src/python/zquantum/core/utils.py
pediejo: Good spot! Fixed this.

src/python/zquantum/core/utils_test.py
pediejo: Okay, tried to add a check of this.

mstechly: Instead of defining your own `SCHEMA_VERSION`, you should do:
`from .utils import SCHEMA_VERSION` :)

PR #72, opened 2020-08-13
Feature/new history recording

dexter2206:
This adds proposed new implementation of storing call history of functions.

### DIFF
--- a/src/python/zquantum/core/interfaces/functions.py
+++ b/src/python/zquantum/core/interfaces/functions.py
@@ -1,5 +1,5 @@
 """Protocols describing different kinds of functions."""
-from typing import NamedTuple, Callable, Any, TypeVar
+from typing import NamedTuple, Callable, Any, TypeVar, Optional
 from typing_extensions import Protocol, runtime_checkable
 import numpy as np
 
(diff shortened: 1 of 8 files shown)


src/python/zquantum/core/interfaces/functions.py
```diff
@@ -0,0 +1,45 @@
+"""Protocols describing different kinds of functions."""
```
mstechly, line 1: I'm not sure whether `cost_function.py` isn't a better place for this? `functions.py` is pretty vague name.
WDYT @simonwa7 ?

src/python/zquantum/core/interfaces/functions.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import NamedTuple, Callable, Any, TypeVar, Optional
+from typing_extensions import Protocol, runtime_checkable
+import numpy as np
+
+T = TypeVar("T", covariant=True)
+S = TypeVar("S", contravariant=True)
+
+
+class StoreArtifact(Protocol):
```
mstechly, line 10: Please add some docstrings @dexter2206 <3

max-radin (approved):

PR #76, opened 2020-08-18
4627013 release pipeline v2

simonwa7:


testing/data_validation/get-bitstring-distribution.py
```diff
@@ -0,0 +1,57 @@
...
+                ]
+                <= 1.0
+            )
+            total_probability += workflowresult[key]["bitstring-distribution"][
+                "bitstring_distribution"
+            ][bitstring]
+
+        print(total_probability)
+        assert total_probability <= 1.0000000000000002
+        assert total_probability >= 0.9999999999999998
```
max-radin, line 54: I'm guessing this is because of some numerical precision issues? It would probably be cleaner to use `np.isclose` instead of less than and greater than.

simonwa7: Addressed here: `b05c1ce`

PR #77, opened 2020-08-18
174266087 initial gate class

simonwa7:


src/python/zquantum/core/circuit/gate/_gate.py
```diff
@@ -0,0 +1,85 @@
...
+        assert self._is_valid_operator(matrix, qubits)
+        assert self._qubits_are_valid(qubits)
+        self.matrix = matrix
+        self.qubits = qubits
+
+    def _is_valid_operator(self, matrix: np.ndarray, qubits: Tuple[int]):
+        # Make sure matrix is square
+        is_square = True
+        num_rows = len(matrix)
+        for row in matrix:
```
mstechly, line 26: This will be more elegant:
`if matrix.shape[0] == matrix.shape[1]` ?

src/python/zquantum/core/circuit/gate/_gate.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    def __init__(self, matrix: np.ndarray, qubits: Tuple[int]):
+        assert self._is_valid_operator(matrix, qubits)
+        assert self._qubits_are_valid(qubits)
+        self.matrix = matrix
+        self.qubits = qubits
+
+    def _is_valid_operator(self, matrix: np.ndarray, qubits: Tuple[int]):
+        # Make sure matrix is square
```
mstechly, line 23: Why are we not checking if matrix is unitary?

src/python/zquantum/core/circuit/gate/_gate.py
```diff
@@ -0,0 +1,85 @@
...
+        for row in matrix:
+            if len(row) != num_rows:
+                is_square = False
+
+        # Make sure matrix is associated with correct number of qubits
+        correct_num_qubits = 2 ** len(qubits) == num_rows
+
+        return is_square and correct_num_qubits
+
+    def _qubits_are_valid(self, qubits: Tuple[int]):
```
mstechly, line 35: If `qubits` is just a tuple of integers, why not check if with `set`?
`if len(set(qubits)) != len(qubits)` ?

src/python/zquantum/core/circuit/gate/_gate_test.py
```diff
@@ -0,0 +1,574 @@
...
+)
+def test_creating_gate_with_repeated_qubits_fails(qubits):
+    """The Gate class should raise an assertion error if all qubit indices are not unique"""
+    matrix = np.eye(2 ** len(qubits))
+    with pytest.raises(AssertionError):
+        Gate(matrix, qubits)
+
+
+@pytest.mark.parametrize("number_of_qubits", [1, 2, 3, 4, 5, 6])
+def test_creating_identity_gate_for_succeeds(number_of_qubits):
```
mstechly, line 72: `test_creating_identity_gate_for_succeeds` -> what does the `for` means here?

src/python/zquantum/core/circuit/gate/_gate_test.py
```diff
@@ -0,0 +1,574 @@
...
+
+
+def test_creating_complex_gate():
+    """The Gate class should be able to handle complex matrices"""
+    # Given
+    qubits = (
+        0,
+        1,
+    )
+    matrix = np.eye(2 ** len(qubits), dtype=complex)
```
mstechly, line 97: I would actually put some actual imaginary/complex numbers to the matrix, not just declare that the type is complex.

src/python/zquantum/core/circuit/gate/_gate_test.py
simonwa7: Was a typo from before I parameterized the test - removed in a later commit

src/python/zquantum/core/circuit/gate/_gate.py
simonwa7: Good idea - was trying to figure out the most efficient way to do this comparison. Added in a future commit

src/python/zquantum/core/circuit/gate/_gate.py
simonwa7: I actually wasn't sure if there would ever be the case someone wanted to use a non-unitary gate - @max-radin @akataba @pediejo thoughts here?

src/python/zquantum/core/circuit/gate/_gate.py
simonwa7: This whole comparison is actually updated in a future commit.

src/python/zquantum/core/circuit/gate/_gate_test.py
simonwa7: Good point - addressed in future commit

src/python/zquantum/core/circuit/gate/_gate.py
max-radin: After discussing with @simonwa7 I think it would make sense to check for unitarity. When modeling nosie, you will have non-unitary operations. But I don't think we need to address this case here.

PR #81, opened 2020-08-28
Build circuitset

martamau:
This PR adds 2 templates: 
1. combine 4 circuit artifacts in a circuitset  
2. add a circuit artifact to an existing circuitset 

Template 2 allows to create bigger circuitsets (more than 4 circuits). 
To me, the ideal way to go here would be to keep in the resource only template 2, and run it in a loop (in the workflow) so to update the circuitset with one circuit at a time, namely: 
```
name: loop-over-circuits
steps:
- - name: add-circuit-artifact-to-circuit-set
     arguments:
     - circuit-set: *the one from the previous loop step*
     - circuit: item 
withItems: [circuit1, circuit2]

```
This seems not to be possible because the input of step n is the output of step n-1, and as far as I know this is not possible to do in the workflow. I looked into argo docs and some examples + thought about some other possible strategies to make this more general but could not find a solution, so I decided to leave the two templates available. Any comments/ideas? 

Context: once you have a circuitset, you can easily combine its elements adding them up to build a more complicated circuit (see [this PR](https://github.com/zapatacomputing/z-quantum-core/pull/82))

### DIFF
--- a/templates/circuit.yaml
+++ b/templates/circuit.yaml
@@ -260,7 +260,7 @@ spec:
           - name: circuit-set
             path: /app/circuit_set.json
 
-    # Add a circuit to a circuit-set 
+    # Add a circuit to a circuit-set
     - name: add-circuit-artifact-to-circuit-set
       parent: generic-task
       inputs:
@@ -269,7 +269,7 @@ spec:
             value: python3 main_script.py
         artifacts:
           - name: circuit-set
-            path: /app/circuit_set.json
+            path: /app/circuit_set_input.json
             optional: true
           - name: circuit
             path: /app/circuit.json
@@ -280,8 +280,8 @@ spec:
                 import os
                 from zquantum.core.circuit import load_circuit_set, load_circuit, save_circuit_set
 
-                if os.path.exists('circuit_set.json'):
-                  circuit_set = load_circuit_set('circuit_set.json')
+                if os.path.exists('circuit_set_input.json'):
+                  circuit_set = load_circuit_set('circuit_set_input.json')
                 else:
                   circuit_set = []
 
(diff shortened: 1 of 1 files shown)


simonwa7:
@martamau I think we can rewrite these templates slightly to join the functionality together.

I like how in `create-circuit-set-from-circuit-artifacts` we can add 4 circuits to create a circuit_set, but if we make circuit-2, circuit-3, and circuit 4 `optional`, then we can create circuit sets without needing _exactly_ 4. 

Then, if we also add `circuit-set` (again as `optional`) we can wrap in the functionality of `add-circuit-artifact-to-circuit-set`.

So the template would look like:
```
if circuit_set.json:
   circuit_set = load_circuit_set("circuit_set.json")
else:
    circuit_set = []

for file in ["circuit_1.json", "circuit_2.json", "circuit_3.json", "circuit_4.json"]:
    if file.exists():
        circuit_set.append(load_circuit(file))

save_circuit_set(circuit_set, "circuit_set.json")
```

martamau:
Yeah, that makes sense. But then in case you want to add e.g. 7 circuits you still have to call the template twice in the workflow, right @simonwa7 ? You would have to build a circuit_set out of the first 4 circuits, and then use this output to add the remaining 3. The ideal would be to have a dynamic number of artifacts but I don't think it's possible?

simonwa7:
Yeah I think that would be the ideal. A good future feature for the platform @Oskoss if it's doable.

martamau:
Ok so in the meantime shall I just go ahead with merging the two templates in one as you suggested and then we'll see what @Oskoss can do for us? :)

simonwa7:
Yep!

martamau:
Done :)

simonwa7 (approved):

PR #82, opened 2020-08-28
adding template to combine circuits

martamau:
This PR adds a template to combine elements of a circuitset to create a composite circuit.

### DIFF
--- a/templates/circuit.yaml
+++ b/templates/circuit.yaml
@@ -229,7 +229,7 @@ spec:
             path: /app/extended_circuit.json
 
     # Combine circuits in a circuitset to create a composite circuit
-    - name: combine-circuits
+    - name: concatenate-circuits
       parent: generic-task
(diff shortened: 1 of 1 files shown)


templates/circuit.yaml
```diff
@@ -227,3 +227,28 @@ spec:
         artifacts:
           - name: extended-circuit
             path: /app/extended_circuit.json
+
+    # Combine circuits in a circuitset to create a composite circuit
+    - name: combine-circuits
```
simonwa7, line 232: Let's rename this to `concatenate-circuits` to avoid potential confustion, but besides that it looks great!

martamau: Addressed in [this commit](https://github.com/zapatacomputing/z-quantum-core/pull/82/commits/e5849cfe0728839d47580356aaa0929c2ec01836)

PR #87, opened 2020-09-07
Evaluate cost function task template

max-radin:
This PR adds a task template to evaluate an ansatz-based cost function for a fixed set of parameters.


steps/measure.py
```diff
@@ -45,3 +50,44 @@ def get_bitstring_distribution(
...
+    qubit_operator: str,
+    noise_model="None",
+    device_connectivity="None",
+):
+    ansatz_parameters = load_circuit_template_params(ansatz_parameters)
+    # Load qubit op
+    operator = load_qubit_operator(qubit_operator)
+    ansatz_specs = json.loads(ansatz_specs)
+    if ansatz_specs["function_name"] == "QAOAFarhiAnsatz":
+        ansatz = create_object(ansatz_specs, cost_hamiltonian=operator)
```
mstechly, line 69: This is not a good solution.
Not sure if this PR is a good place to solve it, but we need to have a story for that.

src/python/zquantum/core/estimator.py
```diff
@@ -97,6 +97,7 @@ def get_estimated_expectation_values(
                     measurements.get_expectation_values(frame_operator)
                 )
             )
+            print(f"Expectation values for {frame_operator}: {expectation_values_set[-1].values}")
```
mstechly, line 100: Do you want to leave it here or is it just some debug code that should disappear?

steps/measure.py
max-radin: Yeah it's not great to hard code behavior for specific ansatzes here. I had just copied this from [optimize_variational_circuit](https://github.com/zapatacomputing/z-quantum-optimizers/blob/master/steps/optimizers.py#L44). Since this pattern is used elsewhere, should we just merge this PR and create a story for figuring out a solution in general?

src/python/zquantum/core/estimator.py
max-radin: I just removed it. We had been wanting this for some experiments but probably better to leave it out so we don't flood the logs.

steps/measure.py
mstechly: Yes, sounds good. Accepting it!

mstechly (approved):

PR #90, opened 2020-09-21
Feature/make value estimate float like

dexter2206:
This makes the `ValueEstimate` a subclass of float. This is another step in modernizing cost function interface. The change is backwards-compatible.


dexter2206:
Something went wrong with diff on GH, closing with the purpose of reopening.

PR #91, opened 2020-09-22
Place finite_differences_gradient to a separate module

dexter2206:
This is the next step towards the integration of new history recording mechanism with optimizers.

The change introduced here decouples cost function from its gradient by introducing gradients module (with factory for finite differences gradient).

Changes presented here are backwards compatible.


src/python/zquantum/core/test_gradients.py
```diff
@@ -0,0 +1,18 @@
...
+    return (parameters ** 2).sum()
+
+
+@pytest.mark.parametrize(
+    "parameters", [np.array([1, 2, 3]), np.array([0.5, 0.25, 1, 0.0])]
+)
+def test_finite_differences_gradient_returns_vectors_with_correct_length(parameters):
+    gradient = finite_differences_gradient(sum_x_squared)
+
+    assert len(parameters) == len(gradient(parameters))
```
mstechly, line 18: Why are we not testing for the values of the gradients?
And whether `epsilon` actually matters?

dexter2206: That's actually a very good question. The reason is: I only ported the tests that were already present in `cost_function_test.py`. But you are right, we can do better. I'll add two more tests checking for the assertions that you mentioned.

dexter2206: @mstechly done!

mstechly (approved):

PR #94, opened 2020-09-24
Reimplement ValueEstimate as a subclass of float

dexter2206:
This reimplements ValueEstimate as a subclass of float. This is backwards compatible, as new `ValueEstimate` still exposes `value` attribute (although marked as deprecated).

### DIFF
--- a/src/python/zquantum/core/utils.py
+++ b/src/python/zquantum/core/utils.py
@@ -1,4 +1,5 @@
 """General-purpose utilities."""
+import warnings
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


mstechly (approved):
Looks good – have you checked if `value` is used somewhere in `core` so it can be fixed already in this commit?

dexter2206:
> Looks good – have you checked if `value` is used somewhere in `core` so it can be fixed already in this PR?

It seems to me that the only place it is in use is in cost functions which we are planning to get rid of anyway.

PR #96, opened 2020-09-29
Adapt optimizer interface to new history recording mechanism

dexter2206:
This adapts optimizer interface to new history recording mechanism (which does not need a `CostFunction` objects). The base class for optimizer tests is also updated so the function being optimized are also callables with gradients as opposed to Cost functions.

### DIFF
--- a/src/python/zquantum/core/interfaces/optimizer_test.py
+++ b/src/python/zquantum/core/interfaces/optimizer_test.py
@@ -1,9 +1,11 @@
 import unittest
 import numpy as np
 import pytest
+from zquantum.core.interfaces.functions import FunctionWithGradient
 
 from .optimizer import optimization_result
 from ..cost_function import BasicCostFunction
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/interfaces/optimizer_test.py
```diff
@@ -35,11 +36,10 @@ def test_optimization(self):
             self.assertIn("opt_params", results.keys())
             self.assertIn("history", results.keys())
 
-    def test_optimization_simple_function(self):
+    def test_optimizer_succeeds_with_optimizing_simple_function(self):
```
mstechly, line 39: I would rename "simple_function" to "sum_of_squares".

dexter2206: Done!

mstechly (approved):

PR #100, opened 2020-10-01
Must load noise model and device connectivity into specs before loadi…

simonwa7:
…ng backend

### DIFF
--- a/steps/measure.py
+++ b/steps/measure.py
@@ -2,10 +2,14 @@ import json
 from zquantum.core.utils import create_object, load_noise_model
 from zquantum.core.circuit import load_circuit, load_circuit_connectivity
 from zquantum.core.bitstring_distribution import save_bitstring_distribution
+from typing import Dict
 
 
 def run_circuit_and_measure(
-    backend_specs, circuit, noise_model="None", device_connectivity="None"
+    backend_specs: Dict,
+    circuit: str,
+    noise_model: str = "None",
+    device_connectivity: str = "None",
 ):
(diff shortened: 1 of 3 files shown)


steps/measure.py
```diff
@@ -7,30 +7,34 @@
...
-        backend_specs_dict["noise_model"] = load_noise_model(noise_model)
+        backend_specs["noise_model"] = load_noise_model(noise_model)
     if device_connectivity != "None":
-        backend_specs_dict["device_connectivity"] = load_circuit_connectivity(
+        backend_specs["device_connectivity"] = load_circuit_connectivity(
             device_connectivity
         )
+
+    backend = create_object(backend_specs)
+    circuit = load_circuit(circuit)
```
mstechly, line 19: `circuit = load_circuit(circuit)` looks confusing. I assume one of these is different from the other?
Also, we're missing typing for these functions.

simonwa7: Added typing here: 8e003d3, the confusing nature of the variable names is because of workflow-v1 -> Can touch base with you about this if needed

mstechly: Can't we rename these variables to "circuit_str" and "circuit_obj" or sth more readable like this?

simonwa7: The variable name is reflected in the workflow, so then we would have `circuit_str` in the workflow which is worse in my opinion.

mstechly: Ok, can we change the name in the python code, add a comment or make it very explicit in the docstrings?

mstechly (approved):

PR #101, opened 2020-10-02
Implement function_with_gradient helper

dexter2206:
This adds the possibility of adding a gradient to functions that store artifacts.

### DIFF
--- a/src/python/zquantum/core/interfaces/functions.py
+++ b/src/python/zquantum/core/interfaces/functions.py
@@ -1,8 +1,9 @@
 """Protocols describing different kinds of functions."""
+from inspect import signature
 from typing import NamedTuple, Callable, Any, TypeVar, Optional, Union
 from typing_extensions import Protocol, runtime_checkable
 import numpy as np
-from zquantum.core.history.recorder import has_store_artifact_param
+
 
 T = TypeVar("T", covariant=True)
 S = TypeVar("S", contravariant=True)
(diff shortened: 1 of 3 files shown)


src/python/zquantum/core/interfaces/functions.py
```diff
@@ -43,10 +45,49 @@ def gradient(self, params: np.ndarray) -> np.ndarray:
...
+class FunctionWithGradientStoringArtifacts(NamedTuple):
+    """A callable with gradient that also stores artifacts."""
+    function: CallableStoringArtifacts
+    gradient: Callable[[np.ndarray], np.ndarray]
+
+    def __call__(self, params: np.ndarray, store_artifact: StoreArtifact):
+        return self.function(params, store_artifact)
+
+
+def function_with_gradient(
```
simonwa7, line 81: Question about the naming convention you're using here:
How come we use CamelCase for `FunctionWithGradient`, yet snake_case for `function_with_gradient`?

dexter2206: Oh, that's because `FunctionWithGradient` is a class and `function_with_gradient` is a function (I guess one could call it a factory for `FunctionWithGradient` and `FunctionWithGradientStoringArtifacts`. That's the same convention as in `recorder.py` where actually the `recorder` is a factory for several of history-recorder classes.

simonwa7 (approved):

PR #102, opened 2020-10-02
Implement serializer and save_optimization_results function

dexter2206:
This implements new `save_optimization_results` function. As a side effect, a more general JSON Encoder implementation is added that should be able to handle most of our serializable objects.

### DIFF
--- a/src/python/zquantum/core/serialization.py
+++ b/src/python/zquantum/core/serialization.py
@@ -18,8 +18,7 @@ def preprocess(tree):
         return {k: preprocess(v) for k, v in tree.items()}
     elif isinstance(tree, tuple) and hasattr(tree, "_asdict"):
         return preprocess(tree._asdict())
-    # Note: isinstance check with ValueEstimate is broken, that's why we compare types here
-    elif type(tree) == ValueEstimate:
+    elif isinstance(tree, ValueEstimate):
         return tree.to_dict()
     elif isinstance(tree, (list, tuple)):
         return list(map(preprocess, tree))
--- a/src/python/zquantum/core/serialization.py
+++ b/src/python/zquantum/core/serialization.py
@@ -1,7 +1,10 @@
 """Serialization module."""
 import json
-from typing import Any, NamedTuple, Iterator
+from operator import attrgetter
+from typing import Any, Iterator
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/serialization.py
```diff
@@ -0,0 +1,48 @@
...
+        return list(map(preprocess, tree))
+    return tree
+
+
+class ZapataEncoder(json.JSONEncoder):
+    ENCODERS_TABLE = {
+        np.ndarray: convert_array_to_dict,
+        ValueEstimate: ValueEstimate.to_dict,
+        BitstringDistribution: attrgetter("distribution_dict")
+    }
```
simonwa7, line 32: This seems like it may be hard to maintain as the list of artifacts grows. Especially in the case of a 3rd party with a custom artifact that they want to be recorded. Perhaps, this problem would be fixed if we begin to standardize the way to serialize artifacts (e.g. a base `Artifact` class that has methods `.to_dict()` that should be implemented), but this should be left for future discussion. Another possibilty would be to make it possible for users to update the `ENCODERS_TABLE` with their own types and methods. 

In the meantime, I'm happy to leave this as is and come back to it at a later date.

dexter2206: I absolutely agree that extensibility of the current solution needs some improvement. Maybe we'll do something like this:

- If an object has `to_dict` method, it will be used for serialization. I don't think we need to have a class hierarchy to do this.
- Otherwise, it would be looked up in the `ENCODERS_TABLE`. This is needed, as we are not able to add `to_dict` method e.g. for numpy arrays.
- For users, we could add a parameter that would allow extending `ENCODERS_TABLE` with custom encoders.

However, I would leave it for now as it is because implementing this will take some time and the current solution seems to be working. Maybe we'll create story in the tracker? What do you think?

simonwa7: That all sounds reasonable to me - I also think that serialization of artifacts is something that @Oskoss and the rest of the platform team has thought about collaborating on, so I'll put a tracker story in and we can touch base with them.

simonwa7 (approved):

PR #103, opened 2020-10-05
Must have schema key in optimization result

simonwa7:


src/python/zquantum/core/serialization_test.py
```diff
@@ -85,3 +87,74 @@ def test_zapata_encoder_can_handle_optimization_result():
...
+    }
+
+    optimization_result_filename = "test-optimization-result-io.json"
+
+    # When
+    save_optimization_results(result_to_serialize, optimization_result_filename)
+
+    # Then
+    with open(optimization_result_filename, "r") as f:
+        loaded_data = json.loads(f.read())
```
dexter2206, line 158: Let's use `json.load(f)` instead of `json.loads(f.read())`.

As a side note, it is not the best idea to use the actual filesystem when unit testing, but I'll explain more about it when we have some time to chat and it would require some rework on our approach for writing serialization functions.

dexter2206 (changes requested):

src/python/zquantum/core/serialization_test.py
simonwa7: Addressed here: 13e53ba

zapatabot (approved):

PR #107, opened 2020-10-16
Fixing indexing of measurements when sampling from wavefunction

simonwa7:


mstechly (approved):
I think we should start being very explicit in the docs as well regarding how we're indexing qubits (if we're not already).

PR #108, opened 2020-10-20
174266170 implement pyquil conversions for gates

dexter2206:
This implements singledispatch-based approach for converting gate objects to PyQuil.

If this gets approved, I'll add dispatch for Circuit object.

Basic idea: there is `convert_to_pyquil` function that dispatches on the type of the object passed and converts it to respective representation in PyQuil.


src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@_convert_gate_to_pyquil.register(CPHASE)
+def convert_CPHASE_to_pyquil(gate: CPHASE) -> pyquil.gates.Gate:
+    return pyquil.gates.CPHASE(gate.angle, *gate.qubits)
+
+
+@_convert_gate_to_pyquil.register(SWAP)
+def convert_SWAP_gate_to_pyquil(gate: SWAP) -> pyquil.gates.Gate:
+    return pyquil.gates.SWAP(*gate.qubits)
```
simonwa7, line 86: This all looks pretty slick, but I have two questions/concerns:
1. pyquil gates can have modifiers that will impact what the gate is intended to do (https://pyquil-docs.rigetti.com/en/stable/basics.html) so we need to check for those
2. This (currently) doesn't work when there's a pyquil gate not in our library or a Zapata gate not in pyquil's library. I think we want to be able to have this flexibility to create custom gates on the fly that won't get lost in translation.

dexter2206:
Closing because there is no way to update diff after I updated base.

PR #109, opened 2020-10-21
173183192 more flexible graphs

FredericSauv:
Changes to produce more flexible distribution of graphs (still working on v0):
 
- New function `generate_random_value_from_string` to generate a random value according to an input string specifying how to produce it.
- `random_weights` arguments can now also be a string defining the distribution from which the weights of the graphs are sampled
- In `generate_graph_from_specs,` some of the values of the specs dictionary can be specified as strings defining how they should be randomly sampled
- Do not regenerate the seed when None is passed
- Replace Cirq gates by their new names (e.g. Rx->rx as per Cirq v0.7.0)
- Require a newer version of Sympy (This was needed for the z-qml image)


src/python/zquantum/core/graph.py
```diff
@@ -119,38 +127,51 @@ def generate_random_regular_graph(
...
     Returns:
         A networkx.Graph object
     """
     assert not(graph.is_multigraph()), "Cannot deal with multigraphs"
-    
-    random.seed(seed)
-    if random_weights:
+    if seed is not None:
+        random.seed(seed)
+    if random_weights == True:
```
mstechly, line 149: Any reasons to add an explicit comparison here?

src/python/zquantum/core/graph.py
```diff
@@ -84,8 +86,11 @@ def generate_random_graph_erdos_renyi(
             Number of nodes.
         probability: float
             Probability of two nodes connecting.
-        random_weights: bool
+        random_weights: bool or str
             Flag indicating whether the weights should be random or constant.
+            By default False, i.e. all the edge weights are set to 1.
```
mstechly, line 91: It's not clear to me why it could be a string.

src/python/zquantum/core/graph.py
```diff
@@ -119,38 +127,51 @@ def generate_random_regular_graph(
...
     Args:
         graph: nx.Graph
             The input graph.
-        random_weights: bool
-            Flag indicating whether the weights should be random or constant (1.).
+        random_weights: bool or str
+            Flag indicating whether the weights should be random or constant.
+            if False (default) static weights, i.e. equal to 1.0
+            if True weights are drawn from uniform(0,1)
+            if a string it will be used when generating random values for the weights
```
mstechly, line 141: It's not specified what would be the format of the string.

mstechly (changes requested):

mstechly:
I don't like the fact that we can pass both `Bool` or `str`. For me it makes the usage of `generate_random_graph_erdos_renyi` just more confusing.
I'd suggest creating a separate function `generate_random_graph_from_string`, so that the code is clearer and we don't try to put everything in one function.

PR #110, opened 2020-10-26
175099277 move agnostic optimization functions to z quantum core

dexter2206:
This implements `load_optimization_results` function capable of decoding the output of `save_optimization_results`.

### DIFF
--- a/src/python/zquantum/core/serialization.py
+++ b/src/python/zquantum/core/serialization.py
@@ -98,6 +98,6 @@ def save_optimization_results(optimization_results, filename):
         json.dump(optimization_results, target_file, cls=OrquestraEncoder)
 
 
-def load_optimization_result(filename):
+def load_optimization_results(filename):
     with open(filename, "rt") as source_file:
         return json.load(source_file, cls=OrquestraDecoder)
--- a/src/python/zquantum/core/serialization.py
+++ b/src/python/zquantum/core/serialization.py
@@ -43,6 +43,24 @@ class ZapataEncoder(json.JSONEncoder):
         return super().iterencode(preprocess(o))
 
 
+class ZapataDecoder(json.JSONDecoder):
+    """Custom decoder for loading data dumped by ZapataEncoder."""
+
+    def __init__(self, *args, **kwargs):
+        json.JSONDecoder.__init__(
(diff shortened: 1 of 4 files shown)


src/python/zquantum/core/serialization.py
```diff
@@ -43,7 +59,45 @@ def iterencode(self, o: Any, _one_shot: bool = ...) -> Iterator[str]:
...
+        "zapata-v1-optimization_result": lambda obj: optimization_result(**obj)
+    }
+
+    def __init__(self, *args, **kwargs):
+        json.JSONDecoder.__init__(
+            self, object_hook=self.object_hook, *args, **kwargs
+        )
+
+    def object_hook(self, obj):
+        # Parts of the below if-elif-else are sketchy, because for some objects there is
```
mstechly, line 76: I'd add a TODO here and appropriate story to the tracker.

mstechly (changes requested):

src/python/zquantum/core/serialization.py
```diff
@@ -3,18 +3,34 @@
...
+from scipy.optimize import OptimizeResult
+
+from .interfaces.optimizer import optimization_result
+from .history.recorder import HistoryEntry, HistoryEntryWithArtifacts
+from .bitstring_distribution import BitstringDistribution, is_bitstring_distribution
 from .utils import convert_array_to_dict, ValueEstimate, SCHEMA_VERSION
 
 
+def has_numerical_keys(dictionary):
+    try:
```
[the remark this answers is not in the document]
dexter2206: This approach is called EAFP, it is outlined e.g. here https://docs.python.org/3/glossary.html

I would typically advocate for using the current approach, but actually, now that you say it I think it is flawed in this particular case because it would also classify strings representing floats as float-like. I fixed it in 091c313

src/python/zquantum/core/serialization.py
dexter2206: It would be the story about overhauling the serialization mechanism we were talking about earlier, right?

mstechly (approved):

src/python/zquantum/core/serialization.py
mstechly: Yeah, that's true 👍

PR #113, opened 2020-11-09
175538989 translate expression from sympy to quil

dexter2206:
This implements converting SymPy expressions into quil-based expressions used in PyQuil, which is a required step for converting our native parametrized gate into PyQuil parameterized gates.

Supported expressions include symbols, numbers, imaginary unit (yes, in SymPy it is not a number :shrug:) known functions that have their counterparts in PyQuil and arithmetic expressions formed by them using usual arithmetic operators, function compositions are also supported.

Additionally, machinery implemented in this PR can serve for implementing other translation layers, which should prove useful when we implement conversion from PyQuil to our native gates.


src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
```diff
@@ -0,0 +1,178 @@
...
+
+@expression_from_sympy.register
+def function_call_from_sympy_function(function: sympy.Function):
+    return FunctionCall(str(function.func), expression_from_sympy(function.args))
+
+
+@expression_from_sympy.register
+def expression_tuple_from_tuple_of_sympy_args(args: tuple):
+    return tuple(expression_from_sympy(arg) for arg in args)
+
```
simonwa7, line 126: Not a must, but would be nice visually to have some separation between `expression_from_sympy` functions and `translate_expression` functions

src/python/zquantum/core/circuit/conversions/symbolic_expressions_test.py
```diff
@@ -0,0 +1,246 @@
...
+    "sympy_expression, quil_expression",
+    [
+        (sympy.Symbol("theta"), quil.Parameter("theta")),
+        (
+            sympy.Mul(sympy.Symbol("theta"), sympy.Symbol("gamma"), evaluate=False),
+            quil.Parameter("theta") * quil.Parameter("gamma"),
+        ),
+        (sympy.cos(sympy.Symbol("theta")), quilatom.quil_cos(quil.Parameter("theta"))),
+        (sympy.cos(2 * sympy.Symbol("theta")), quilatom.quil_cos(2 * quil.Parameter("theta")))
+    ]
```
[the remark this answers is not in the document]
dexter2206: Great idea. I added some tests in 0157e49
I had to add some additional detection of sqrt because SymPy converts it to pow(..., 0.5), but it should work no and the tests were added here: f5de537
There is no `quil_tan` function but I defined it in `QUIL_DIALECT` using sine and cosine. Implementation and tests were added here: c71e207

Note that some new tests cases once again use e.g. `sympy.Add` explicitly, that's because it's quite possible that otherwise SymPy would massacre the order of arguments (resulting in equivalent, but not an equal expression).

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
```diff
@@ -0,0 +1,178 @@
...
+        "add": operator.add,
+        "mul": operator.mul,
+        "div": operator.truediv,
+        "sub": operator.sub,
+        "cos": quilatom.quil_cos,
+        "sin": quilatom.quil_cos,
+        "exp": quilatom.quil_exp,
+        "sqrt": quilatom.quil_sqrt
+    }
+)
```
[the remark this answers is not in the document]
dexter2206: The dialect objects are meant to live here long-term. When implementing other conversions we just need to add new dialect objects. That way the implementation of the translation process (at least from SymPy to something else) stays the same and new targets can be declared by only adding new dialects.

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class FunctionCall(NamedTuple):
+    """Represents abstract function call.     """
+
+    name: str
+    args: Iterable["Expression"]
+
+
+Expression = Union[Symbol, FunctionCall, Number]
```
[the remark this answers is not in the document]
dexter2206: I think those are the only possibilities. Anyway, if we'll find later that there's some category of expression tree node missing, we'll just add it here along with a dispatch to construct it from SymPy expression.

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
dexter2206: You are right, it would be nice to separate them somehow. However, I'm not entirely convinced about the readability of including some visual markers in form of comments or whitespaces. What do you think about putting those in separate modules?

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def translate_tuple(expression_tuple: Iterable[Expression], dialect: ExpressionDialect):
+    return tuple(
+        translate_expression(element, dialect)
+        for element in expression_tuple
+    )
+
+
+QUIL_DIALECT = ExpressionDialect(
```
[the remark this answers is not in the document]
dexter2206: Added a comment in 77e4153

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
simonwa7: I like this idea best. Maybe split it so that we have:
```
- conversions

File: src/python/zquantum/core/circuit/conversions/symbolic_expressions_test.py [hunk 6]
```diff
@@ -268,6 +268,22 @@ def test_sympy_function_calls_are_converted_to_function_call_object_with_appropr
...
+            sympy.Symbol("y") ** sympy.Symbol("x"),
+            quil.Parameter("y") ** quil.Parameter("x")
+        ),
+        (
+            sympy.Symbol("x") ** 2,
+            quil.Parameter("x") ** 2
+        ),
+        (
+            2 ** sympy.Symbol("x"),
+            2 ** quil.Parameter("x")
```
simonwa7, line 286: This seems to be a repeat of 273->274 no?

src/python/zquantum/core/circuit/conversions/symbolic_expressions.py
dexter2206: I suggested something similar to what you suggested in 392f3a3 and 51e8fa5.

src/python/zquantum/core/circuit/conversions/symbolic_expressions_test.py
dexter2206: Nive catch. I removed the duplicate.

PR #115, opened 2020-11-11
Updated installation instructions in readme.

mstechly:


README.md
```diff
@@ -65,7 +65,7 @@ save_circuit(circuit, 'circuit.json')
 ```
 
 Even though it's intended to be used with Orquestra, `z-quantum-core` can be also used as a standalone Python module.
-This can be done by running `pip install .` from the `src/` directory.
+This can be done by running `pip install -e .` from the `src/` directory.
```
simonwa7, line 68: I would add that it can also be done by `pip install .` from the root directory. Especially since I think we'll be moving toward supporting only the root `setup.py` in the future

simonwa7 (approved):

PR #116, opened 2020-11-11
Basic estimator now accounts for constant term

max-radin:


mstechly:
Overall looks good, but I would add a test for a term which contains both constant and non-constant term.
Just in case.

src/python/zquantum/core/estimator_test.py
```diff
@@ -119,6 +119,26 @@ def test_n_samples_is_restored(self):
             # Then
             self.assertEqual(self.backend.n_samples, 5)
 
+    def test_get_estimated_expectation_values(self):
+        for estimator in self.estimators:
+            # Given
+            coefficient = -2
+            constant_qubit_operator = QubitOperator((), coefficient)
```
mstechly, line 126: I would change it to the operator having both constant and non-constant terms.

mstechly (changes requested):

mstechly (approved):

PR #118, opened 2020-11-16
Removing classical and quantum registers from to/from qiskit function…

simonwa7:
…s. All qubits will be added to the same quantum register

### DIFF
--- a/src/python/zquantum/core/circuit/_gate.py
+++ b/src/python/zquantum/core/circuit/_gate.py
@@ -34,7 +34,7 @@ class Gate(object):
         params: list[any]
             List of parameters associated with the gate.
         The above three attributes follow a standard convention that is enforced throughout
-        Zap OS.
+        z-quantum-core
 
         (optional attribute)
         info: dictionary
--- a/src/python/zquantum/core/circuit/_gate.py
+++ b/src/python/zquantum/core/circuit/_gate.py
@@ -23,18 +23,18 @@ from ._qubit import Qubit
 
 class Gate(object):
     """Class for storing information associated with a quantum gate.
- 
+
     Attributes:
         name: string
             Name of the gate. Examples include 'H', 'CNOT' etc.
             The validity of the name goes hand in hand with the specific package
             for quantum computation, as described in 'label'.
         qubits: list[Qubit]
-            List of core.qubit.Qubit object that the gate operates on. 
+            List of core.qubit.Qubit object that the gate operates on.
         params: list[any]
-            List of parameters associated with the gate. 
+            List of parameters associated with the gate.
         The above three attributes follow a standard convention that is enforced throughout
-        Zap OS. 
+        Zap OS.
 
         (optional attribute)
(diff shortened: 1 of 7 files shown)


src/python/zquantum/core/circuit/_gate.py
```diff
@@ -23,18 +23,18 @@
...
             for quantum computation, as described in 'label'.
         qubits: list[Qubit]
-            List of core.qubit.Qubit object that the gate operates on. 
+            List of core.qubit.Qubit object that the gate operates on.
         params: list[any]
-            List of parameters associated with the gate. 
+            List of parameters associated with the gate.
         The above three attributes follow a standard convention that is enforced throughout
-        Zap OS. 
+        Zap OS.
```
mstechly, line 37: I think we can drop "Zap OS" naming from here ;)

src/python/zquantum/core/circuit/_gate.py
```diff
@@ -377,7 +386,7 @@ def to_cirq(self, input_cirq_qubits=None):
         if self.name == "ZZ":
             return cirq.ZZ(cirq_qubits[0], cirq_qubits[1]) ** (params[0] * 2 / pi)
 
-    def to_qiskit(self, qreg=None):
+    def to_qiskit(self, qreg, creg):
```
mstechly, line 389: Shouldn't `creg` be an optional parameter that we should create as some default if it's not provided?

src/python/zquantum/core/utils_test.py
```diff
@@ -304,19 +308,10 @@ def test_two_value_estimates_are_equal_iff_their_values_and_precisions_are_equal
 
 
 @pytest.mark.parametrize(
-    "estimate",
-    [
-        ValueEstimate(2.0),
-        ValueEstimate(5.0, precision=1e5)
-    ]
+    "estimate", [ValueEstimate(2.0), ValueEstimate(5.0, precision=1e5)]
```
mstechly, line 311: It doesn't seem to really matter, but precision here is `1e5`, shouldn't it be `1e-5`?

mstechly (approved):

src/python/zquantum/core/circuit/_gate.py
simonwa7: I was toying with this idea, but I don't think it makes sense to create registers that are only "seen" by the Gate class. The register should be created/controlled by the circuit itself.

src/python/zquantum/core/circuit/_gate.py
simonwa7: bc7736d

src/python/zquantum/core/utils_test.py
simonwa7: ccf2ad5

src/python/zquantum/core/circuit/_gate.py
mstechly: Ok, sounds good to me :)

PR #119, opened 2020-11-17
175538989 add support for converting parametrized gates

dexter2206:
This is a (hopefully) final step in our efforts to implement the conversion of arbitrary Orquestra gates to PyQuil gates.

At this stage, it should be possible to convert an Orquestra gate to PyQuil gate even if it contains symbolic parameters. The symbolic parameters are also collected, i.e. appropriate variable declarations are also added to the corresponding PyQuil program in which context's the conversion is happening.


dexter2206:
Deleting because I used the wrong target branch :(

PR #120, opened 2020-11-17
175538989 add support for converting parametrized gates

dexter2206:
This is a (hopefully) final step in our efforts to implement the conversion of arbitrary Orquestra gates to PyQuil gates.

At this stage, it should be possible to convert an Orquestra gate to PyQuil gate even if it contains symbolic parameters. The symbolic parameters are also collected, i.e. appropriate variable declarations are also added to the corresponding PyQuil program in which context's the conversion is happening.

### DIFF
--- a/src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
+++ b/src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
@@ -94,10 +94,7 @@ def test_converting_rotation_gate_to_pyquil_preserves_qubit_index_and_angle(
 
 
 @pytest.mark.parametrize("qubit", [0, 4, 10, 11])
-@pytest.mark.parametrize(
-    "zquantum_angle, pyquil_angle",
-    EXAMPLE_PARAMETRIZED_ANGLES
-)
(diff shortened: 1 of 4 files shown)


mstechly (approved):

simonwa7 (commented):
@dexter2206 I think this all looks great! So much nicer and cleaner than what we have today. 

One question I have is if we convert from pyquil -> zapata -> pyquil or zapata -> pyquil -> zapata, do we retain all the same information? If so, we should try to test these two cases.

The only other thing I would add is we should add a simple test case where we pass in an empty circuit and try to do the conversion and make sure everything works as expected.

dexter2206:
@simonwa7 
> One question I have is if we convert from pyquil -> zapata -> pyquil or zapata -> pyquil -> zapata, do we retain all the same information? If so, we should try to test these two cases.

We don't have pyquil -> zapata conversion yet, but I can already tell you that this won't work in some cases when modifiers are used. The reason is, that in our gates we optimize modifiers so e.g. X(0).dagger or X(0).dagger.dagger is just X(0) because X is hermitian, but this is not the case in PyQuil (which makes sense, such optimizations are probably left to Quil compiler).

> The only other thing I would add is we should add a simple test case where we pass in an empty circuit and try to do the conversion and make sure everything works as expected.

Good idea, I added one in 680e84b.

PR #125, opened 2020-11-30
Added handling numpy values when serializing measurements.

mstechly:


simonwa7:
Quick question @mstechly, should the casting to `int` occur when you initialize the measurements/append to the list of bitstrings?

mstechly:
I don't think so – it would be an unnecessary overhead that wouldn't necessarily be helpful at all (since not all `Measurements` instances will end up being serialized).

simonwa7 (approved):

PR #128, opened 2020-12-02
Measurement analysis for QubitOperators

aakunitsa:
The branch implements measurement analysis for QubitOperator-s with greedy grouping and (optionally) expectation values of Pauli strings extracted from InteractionRDM-s. I have also added serialization/deserialization functions for "measurement estimates" and auxiliary functions in `steps` so the new features can be used in v1 workflows.


src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,124 @@ def group_comeasureable_terms_greedy(
...
+            offset = 0 if i == 0 else np.sum(group_sizes[:i])
+            pauli_variances_for_group = pauli_variances[offset : offset + group_sizes[i]]
+            frame_variances.append(np.sum(coeffs**2 * pauli_variances_for_group))
+
+    return np.array(frame_variances)
+
+def get_expectation_values_from_rdms(interactionrdm: InteractionRDM, 
+        qubitoperator: QubitOperator, 
+        sort_terms: bool = False) -> ExpectationValues:
+    """ Computes expectation values of a qubitOperator
```
dexter2206, line 108: The docstring here is inconsistent with previous docstrings in terms of indentation. It would be great if we could uniformize this.

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,124 @@ def group_comeasureable_terms_greedy(
...
+
+    expectations_packed = interactionrdm.get_qubit_expectations(reordered_qubitoperator)
+
+    if () in expectations_packed.terms:
+        del expectations_packed.terms[()] # Expectation of the constant term is excluded from expectation values
+
+    expectations = np.real(np.array(list(expectations_packed.terms.values()))) # should we add an assert to catch large Im parts
+    # Clip expectations if they fell out of [-1 , 1] due to numerical errors
+    expectations[expectations < -1.0] = -1.0
+    expectations[expectations >  1.0] =  1.0
```
dexter2206, line 139: I don't know if you are familiar with it, but there's `np.clip` function that does precisely that but can handle both edges of the interval at the same time. I think the usage here would look like this: `np.clip(expectations, -1, 1, out=expectations)`. Using `out` keyword will perform the operation in place, which would be consistent with the current code.

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,124 @@ def group_comeasureable_terms_greedy(
...
+    elif decomposition_method == 'greedy':
+        decomposition_function = lambda qubit_operator: group_comeasureable_terms_greedy(
+        qubit_operator, False)
+    else:
+        raise Exception(f'{decomposition_method} grouping is not implemented')
+
+    groups = decomposition_function(target_operator)
+    frame_variances = compute_group_variances(groups, expecval)
+    sqrt_lambda = sum(np.sqrt(frame_variances))
+    frame_meas = np.asarray([sqrt_lambda * np.sqrt(x) for x in frame_variances])
```
dexter2206, line 190: `compute_group_variances` returns an array, so we could use broadcasting here: 
`frame_meas = sqrt_lambda * np.sqrt(frame_variances)`

src/python/zquantum/core/hamiltonian_test.py
```diff
@@ -54,3 +166,115 @@ def test_group_comeasureable_terms_greedy_sorted(
...
+        coeff = np.array(list(qubitoperator.terms.values()))
+        coeff = coeff[1:]
+    else:
+        sorted_qubitoperator = QubitOperator((), 0.0)
+        terms_iterator = sorted(
+            qubitoperator.terms.items(), key=lambda x: abs(x[1]), reverse=True
+        )
+        coeff = []
+        for term, coefficient in terms_iterator:
+            if not term == ():
```
dexter2206, line 198: It is more Pythonic to use `!=` operator instead of combination of `not` and `==`, i.e. `if term != ()`.

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,124 @@ def group_comeasureable_terms_greedy(
...
+
+    frame_variances = None
+    if decomposition_method == 'greedy-sorted':
+        decomposition_function = lambda qubit_operator: group_comeasureable_terms_greedy(
+        qubit_operator, True) 
+    elif decomposition_method == 'greedy':
+        decomposition_function = lambda qubit_operator: group_comeasureable_terms_greedy(
+        qubit_operator, False)
+    else:
+        raise Exception(f'{decomposition_method} grouping is not implemented')
```
dexter2206, line 185: We should probably raise here `ValueError` instead of `Exception`, which would make it easier to catch and handle if needed.

steps/measure.py
```diff
@@ -91,3 +93,29 @@ def evaluate_ansatz_based_cost_function(
...
     save_value_estimate(value_estimate, "value_estimate.json")
+
+def hamiltonian_analysis(
+    qubit_operator: str,
+    decomposition_method: str = "greedy",
+    expectation_values: str = "None",
+):
+    operator = load_qubit_operator(qubit_operator)
+    if decomposition_method != "greedy-sorted" and decomposition_method != "greedy":
+        raise Exception(f'Decomposition method {decomposition_method} is not supported')
```
dexter2206, line 104: As previously, we should probably raise `ValueError` here.

steps/measure.py
```diff
@@ -91,3 +93,29 @@ def evaluate_ansatz_based_cost_function(
     value_estimate = cost_function(ansatz_parameters)
 
     save_value_estimate(value_estimate, "value_estimate.json")
+
+def hamiltonian_analysis(
+    qubit_operator: str,
+    decomposition_method: str = "greedy",
+    expectation_values: str = "None",
```
dexter2206, line 100: Here we use "None" as the default value but the only way it is used is to convert it to `None` object later. Maybe we could just use `None` as the default value and skip `else` in one of the conditional expressions below?

dexter2206 (changes requested):
From the implementation perspective, this looks good, I only left minor suggestions.

However, I think that there is still work to be done on docstring formatting. Also, I think black should be applied to the code before it can be merged, so the code formatting is consistent with rest of the codebase.

For docstrings, you might want to look at the example here to see the correct indentation: https://google.github.io/styleguide/pyguide.html#doc-function-raises

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,124 @@ def group_comeasureable_terms_greedy(
             groups.append(QubitOperator(term, qubit_operator.terms[term]))
 
     return groups
+
+def compute_group_variances(groups: List[QubitOperator], expecval: ExpectationValues=None) -> np.array:
+    """Computes the variances of each frame in a grouped operator.
```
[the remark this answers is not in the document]
aakunitsa: Thanks for pointing this out! I have addressed this issue in [f14a457](https://github.com/zapatacomputing/z-quantum-core/pull/128/commits/f14a457c472a8c72d701c5e1c1c9bbcd175c4403) Hopefully, it looks good now!

src/python/zquantum/core/hamiltonian.py
aakunitsa: Sorry about that! I changed the indentation in the latest commit.

src/python/zquantum/core/hamiltonian.py
aakunitsa: I replaced my code with the `np.clip` call (I never used it before, it's handy! Thank you!)

src/python/zquantum/core/hamiltonian.py
aakunitsa: 100%! I just copypasted it from the old zmachine code :p. Thanks for spotting this!

src/python/zquantum/core/hamiltonian_test.py
aakunitsa: Fixed!

steps/measure.py
aakunitsa: Fair point, I replaced `Exception` with `ValueError` in both cases

steps/measure.py
aakunitsa: I have replaced "None" and removed `else` but I am not sure if that would cause any issues in V1 workflows. Other functions in `measure.py` use "None"-s so I tried to be consistent in my original implementation.

aakunitsa:
Thanks a lot for your comments, Konrad! I have never used black. Are there any specific command-line options that I should set when running it?

jgonthier:
> Thanks a lot for your comments, Konrad! I have never used black. Are there any specific command-line options that I should set when running it?

You can install black in VSCode and have it format automatically on save: https://marcobelo.medium.com/setting-up-python-black-on-visual-studio-code-5318eba4cd00
Not quite sure about command line usage though.

jgonthier (approved):
Approved with minor revisions :D

dexter2206:
> > Thanks a lot for your comments, Konrad! I have never used black. Are there any specific command-line options that I should set when running it?
> 
> You can install black in VSCode and have it format automatically on save: https://marcobelo.medium.com/setting-up-python-black-on-visual-studio-code-5318eba4cd00
> Not quite sure about command line usage though.

Command line usage is really simple, just `black <path>` and it will basically reformat all files in place. However, integrating it with your IDE is probably much better option.

dexter2206 (approved):
Looks good, let's get it merged once remarks from other reviewers are addressed.

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,125 @@ def group_comeasureable_terms_greedy(
...
+    reordered_qubitoperator = QubitOperator()
+    for term, coefficient in terms_iterator:
+        reordered_qubitoperator += QubitOperator(term, coefficient)
+
+    expectations_packed = interactionrdm.get_qubit_expectations(reordered_qubitoperator)
+
+    if () in expectations_packed.terms:
+        del expectations_packed.terms[()] # Expectation of the constant term is excluded from expectation values
+
+    expectations = np.real(np.array(list(expectations_packed.terms.values()))) # should we add an assert to catch large Im parts
```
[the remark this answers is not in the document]
aakunitsa: I think there is a function in `z-quantum-core` that can be used to convert the complex expectation values into real, but it does not generate any exceptions if it encounters large imaginary parts. I'll switch to it anyway to make the code more maintainable in the future. Also, I'll add an exception for imaginary parts > 10^-3

src/python/zquantum/core/hamiltonian.py
```diff
@@ -69,3 +71,125 @@ def group_comeasureable_terms_greedy(
...
+    else:
+        terms_iterator = qubitoperator.terms.items()
+    reordered_qubitoperator = QubitOperator()
+    for term, coefficient in terms_iterator:
+        reordered_qubitoperator += QubitOperator(term, coefficient)
+
+    expectations_packed = interactionrdm.get_qubit_expectations(reordered_qubitoperator)
+
+    if () in expectations_packed.terms:
+        del expectations_packed.terms[()] # Expectation of the constant term is excluded from expectation values
```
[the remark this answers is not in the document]
aakunitsa: I have eliminated the constant term in `compute_group_variances` as well. The same convention was originally adopted in `get_estimated_expectation_values` for the estimators. It does get appended at the end after concatenating all the measurement sets but I think it's not a very good design choice since we don't have a way to tell whether we included it or not without knowing the initial qubit operator. I'd be curious to know what @max-radin thinks.

aakunitsa:
Thanks a lot, @jgonthier and @dexter2206! (also, black rocks!)

aakunitsa:
@dexter2206, sorry, but I had to revert the change from "None" to None in `steps/measure.py` to make the `hamiltonian_analysis` function compatible with V1. The current version of the workflow language does not allow to omit optional step parameters and does not seem to allow None values.  Could you please have a look one more time to make sure that it looks good?

dexter2206:
> @dexter2206, sorry, but I had to revert the change from "None" to None in `steps/measure.py` to make the `hamiltonian_analysis` function compatible with V1. The current version of the workflow language does not allow to omit optional step parameters and does not seem to allow None values. Could you please have a look one more time to make sure that it looks good?

Sorry for suggesting the change, I am not that familiar with workflow language yet.

dexter2206 (approved):

aakunitsa:
No problem, replacing "None" with `None` is perfectly reasonable! I hope we'll do it at some point when the workflow language gets more mature! :) Thanks a lot for approving the changes!

PR #130, opened 2020-12-03
175540559 implement conversion from pyquil to orquestra

dexter2206:
This adds conversion of basic gates from PyQuil to Orquestra and simplifies already present conversion tests.

Changes from this PR don't include

- Conversion of custom gates
- Conversion of a whole program
- Proper docstrings

The above mentioned missing parts will be added in the next PR.

### DIFF
--- a/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
+++ b/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
@@ -1,6 +1,5 @@
 """Utilities for converting gates and circuits to and from Pyquil objects."""
 from functools import singledispatch
-from itertools import chain
 from typing import Union, Optional, overload, Iterable
 import numpy as np
 import pyquil
--- a/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
+++ b/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
@@ -230,24 +230,43 @@ def convert_from_pyquil(obj: Union[pyquil.Program, pyquil.quil.Gate]):
 
 @convert_from_pyquil.register
 def convert_gate_from_pyquil(gate: pyquil.quil.Gate) -> Gate:
+    number_of_control_modifiers = gate.modifiers.count("CONTROLLED")
+
+    all_qubits = pyquil_qubits_to_numbers(gate.qubits)
+
+    control_qubits = all_qubits[:number_of_control_modifiers]
+    original_qubits = all_qubits[number_of_control_modifiers:]
+
     try:
         gate_cls = PYQUIL_NAME_TO_ORQUESTRA_CLS[gate.name]
         if (
             gate_cls in SINGLE_QUBIT_NONPARAMETRIC_GATES
             or gate_cls in TWO_QUBIT_CONTROLLED_GATES
         ):
-            return gate_cls(*pyquil_qubits_to_numbers(gate.qubits))
+            result = gate_cls(*original_qubits)
         elif gate_cls in ROTATION_GATES or gate_cls == CPHASE:
-            return gate_cls(
-                *pyquil_qubits_to_numbers(gate.qubits),
+            result = gate_cls(
+                *original_qubits,
                 translate_expression(
                     expression_from_pyquil(gate.params[0]),
                     SYMPY_DIALECT
                 )
             )
         elif gate_cls == SWAP:
-            return gate_cls(pyquil_qubits_to_numbers(gate.qubits))
+            result = gate_cls(original_qubits)
+        else:
+            raise RuntimeError(
+                f"Error converting gate {gate}. If you see this message, "
+                "please file a bugreport."
+            )
 
+        for qubit in reversed(control_qubits):
+            result = ControlledGate(result, qubit)
+
+        if gate.modifiers.count("DAGGER") % 2 == 1:
+            result = result.dagger
+
+        return result
     except KeyError:
         raise ValueError(
             f"Conversion to Orquestra is not supported for {gate.name} gate"
--- a/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
+++ b/src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
@@ -52,11 +52,10 @@ ROTATION_GATES = {
 TWO_QUBIT_CONTROLLED_GATES = {
     CZ: pyquil.gates.CZ,
     CNOT: pyquil.gates.CNOT,
-    SWAP: pyquil.gates.SWAP,
 }
 
 
-PYQUIL_NAME_TO_ORQUESTRA_NAME = {
+PYQUIL_NAME_TO_ORQUESTRA_CLS = {
     cls.__name__: cls
     for cls in chain(
         SINGLE_QUBIT_NONPARAMETRIC_GATES,
@@ -64,6 +63,9 @@ PYQUIL_NAME_TO_ORQUESTRA_NAME = {
         ROTATION_GATES
     )
 }
+
+PYQUIL_NAME_TO_ORQUESTRA_CLS["CPHASE"] = CPHASE
+PYQUIL_NAME_TO_ORQUESTRA_CLS["SWAP"] = SWAP
 
 
 def pyquil_qubits_to_numbers(qubits: Iterable[pyquil.quil.Qubit]):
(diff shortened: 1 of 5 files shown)


src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+    pass
+
+
+@convert_from_pyquil.register
+def convert_gate_from_pyquil(gate: pyquil.quil.Gate) -> Gate:
+    number_of_control_modifiers = gate.modifiers.count("CONTROLLED")
+
+    all_qubits = pyquil_qubits_to_numbers(gate.qubits)
+    control_qubits = all_qubits[:number_of_control_modifiers]
+    original_qubits = all_qubits[number_of_control_modifiers:]
```
simonwa7, line 237: Does "original" here actually mean "target"?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
         program += convert_to_pyquil(gate, program)
 
     return program
+
+
+@singledispatch
+def convert_from_pyquil(obj: Union[pyquil.Program, pyquil.quil.Gate]):
+    pass
```
simonwa7, line 228: I know this isn't finished yet, but would it be better here to raise an `Exception` by default?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -50,18 +52,32 @@
 TWO_QUBIT_CONTROLLED_GATES = {
     CZ: pyquil.gates.CZ,
     CNOT: pyquil.gates.CNOT,
-    SWAP: pyquil.gates.SWAP,
 }
 
 
+PYQUIL_NAME_TO_ORQUESTRA_CLS = {
```
simonwa7, line 58: I'm assuming this is meant to just be a "reversed" mapping compared to `SINGLE_QUBIT_NONPARAMETRIC_GATES`, `ROTATION_GATES`, and `TWO_QUBIT_CONTROLLED_GATES`? If so, let's just add a comment here explaining the relationship.

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -50,18 +52,32 @@
...
+    cls.__name__: cls
+    for cls in chain(
+        SINGLE_QUBIT_NONPARAMETRIC_GATES, TWO_QUBIT_CONTROLLED_GATES, ROTATION_GATES
+    )
+}
+
+PYQUIL_NAME_TO_ORQUESTRA_CLS["CPHASE"] = CPHASE
+PYQUIL_NAME_TO_ORQUESTRA_CLS["SWAP"] = SWAP
```
simonwa7, line 66: These seem like outliers and I think it's fine for now, but it's probably unsustainable in the long-term to have these act as "special cases". Maybe, for the future, we can think about better ways to describe/categorize gates so that we run into these cases less frequently.

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+    original_qubits = all_qubits[number_of_control_modifiers:]
+
+    try:
+        gate_cls = PYQUIL_NAME_TO_ORQUESTRA_CLS[gate.name]
+        if (
+            gate_cls in SINGLE_QUBIT_NONPARAMETRIC_GATES
+            or gate_cls in TWO_QUBIT_CONTROLLED_GATES
+        ):
+            result = gate_cls(*original_qubits)
+        elif gate_cls in ROTATION_GATES or gate_cls == CPHASE:
```
simonwa7, line 246: Noting comment above about having `CPHASE` and `SWAP` being special cases might be unsustainable long-term

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+            result = gate_cls(
+                *original_qubits,
+                translate_expression(
+                    expression_from_pyquil(gate.params[0]), SYMPY_DIALECT
+                ),
+            )
+        elif gate_cls == SWAP:
```
simonwa7, line 253: Just noting the special case again

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
simonwa7: Also, I feel like saying `CPHASE` isn't a "controlled two qubit gate" isn't true. Maybe best to rename `TWO_QUBIT_CONTROLLED_GATES` to `TWO_QUBIT_NONPARAMETRIC_CONTROLLED_GATES`?

simonwa7: A possible longer-term solution would be to have a map from gate name to a tuple of `(number_of_qubits: int, is_parametric: bool, is_controlled: bool)` and then build logic off of that knowledge?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+            result = gate_cls(original_qubits)
```
simonwa7, line 254: This may be something I'm just unaware of, but why are you passing `*original_qubits` in the other cases, but `original_qubits` here?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+        else:
+            raise RuntimeError(
+                f"Error converting gate {gate}. If you see this message, "
+                "please file a bugreport."
+            )
+
+        for qubit in reversed(control_qubits):
```
simonwa7, line 261: Please add a comment here explaining why we flip the controlled qubits

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -203,3 +221,51 @@ def convert_circuit_to_pyquil(
...
+            result = ControlledGate(result, qubit)
+
+        if gate.modifiers.count("DAGGER") % 2 == 1:
```
simonwa7, line 264: This seems odd. Does `pyquil` support doing multiple `DAGGER` modifiers as opposed to just a toggle?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: Original here refers to qubits that the gate acted on before applying modifiers. I agree that `target_qubits` is a better name for this variable. Fixed in 0ef10bb

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: Great idea! Fixed in 2d8f706

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: I like the idea of building logic based on the map describing properties of gates, but see comment below about the SWAP gate. Also, you are right that `TWO_QUBIT_NONPARAMETRIC_CONTROLLED_GATES` is a better name, I changed it in 8781468 and in 2a7a4e8 I added comment about the purpose of  `PYQUIL_NAME_TO_ORQUESTRA_CLS`.

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: This is because SWAP gate has different initializer than other two-qubit gates. I think the signatures were already present in ddc4ce35ac57473eb4c811ce33105d46a27d7024 and I didn't recognize a need for uniformizing the signatures until this very point. This difference in signatures also prevents us from implementing this better logic that you suggested above. What do you think about changing the signature of SWAP gate's init so that it matches other two-qubit gates?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: Comment added in deadd59

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: Unfortunately yes, PyQuil allows applying multiple dagger modifiers. I suppose that optimizing them out is left to the Quil compiler later on so they didn't bother to implement it in PyQuil itself. To see that this is the case, consider the following code, which prints `["DAGGER", "DAGGER", "DAGGER"]`
```python
from pyquil.gates import X
gate = X(2).dagger().dagger().dagger()
print(gate.modifiers)
```

dexter2206: I added a comment in d82c2a1 that describes why we test for number of DAGGER modifiers.

simonwa7 (approved):

PR #131, opened 2020-12-08
175540559 converting pyquil custom gates and programs

dexter2206:
The final step of conversion between PyQuil objects and orquestra objects. `convert_from_pyquil` variant implemented here allows for converting whole PyQuil program, including ones with custom gates.

### DIFF
--- a/src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
+++ b/src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
@@ -71,6 +71,83 @@ def pyquil_gate_matrix(gate: pyquil.gates.Gate) -> np.ndarray:
         return QUANTUM_GATES[gate.name]
 
 
+@pytest.fixture
+def example_orquestra_circuit():
+    """Constructs an example_orquestra_circuit.
+
+    This circuit should be equivalent to the the one constructed by
+    the example_pyquil_circuit. This equivalence is used in
+    conversion tests in both ways.
+    """
+    theta = sympy.Symbol("theta")
+
+    return Circuit(
+        [
+            CustomGate(
+                sympy.Matrix(
+                    [
+                        [sympy.cos(theta), sympy.sin(theta)],
+                        [-sympy.sin(theta), sympy.cos(theta)],
+                    ]
+                ),
+                (0,),
+                name="V",
+            ),
+            X(0),
+            Y(1).dagger,
+            Z(3),
+            ControlledGate(SWAP(0, 2), 1),
+            CustomGate(
+                sympy.Matrix(
+                    [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, -1j], [0, 0, -1j, 0]]
+                ),
+                (1, 3),
+                name="U",
+            ).dagger,
+            RX(2, np.pi),
+            CNOT(1, 3),
+        ]
+    )
+
+
+@pytest.fixture
+def example_pyquil_program():
+    quil_theta = pyquil.quil.Parameter("theta")
+
+    u_gate_definition = pyquil.quil.DefGate(
+        "U", [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, -1j], [0, 0, -1j, 0]]
+    )
+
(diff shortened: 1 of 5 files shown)


src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
```diff
@@ -376,6 +453,97 @@ def test_dagger_of_orquestra_gate_is_taken_when_converting_gate_with_dagger_modi
...
+                (3,)
+            ),
+            (
+                "V",
+                [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, -1j], [0, 0, -1j, 0]],
+                (1, 4)
+            )
+        ]
+    )
+    def test_factory_created_form_nonparametrized_pyquil_defgate_produces_correct_gates(
```
simonwa7, line 472: `form` -> `from`?

src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
```diff
@@ -472,81 +640,31 @@ def test_passing_manually_constructed_gate_with_mismatching_properties_raises_va
...
 
 
-def test_converting_circuit_to_pyquil_gives_program_with_the_same_gates():
-    # The goal of the program constructed below is to include a diverse range
-    # of gates.
-    theta = sympy.Symbol("theta")
-    quil_theta = pyquil.quil.Parameter("theta")
+def test_converting_circuit_to_pyquil_gives_program_with_the_same_gates(
+    example_orquestra_circuit, example_pyquil_program
+):
```
simonwa7, line 645: I might just be missing something simple, but does this function actually get called during testing? And if so, where are the inputs for `example_orquestra_circuit` and `example_pyquil_program` defined?

simonwa7: Or perhaps this is to prepare for future testing?

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
```diff
@@ -231,14 +235,54 @@ def convert_circuit_to_pyquil(
...
 @singledispatch
-def convert_from_pyquil(obj: Union[pyquil.Program, pyquil.quil.Gate]):
+def convert_from_pyquil(obj: Union[pyquil.Program, pyquil.quil.Gate], custom_gates=None):
     raise NotImplementedError(
         f"Conversion from pyquil to orquestra not implemented for {obj}"
     )
 
 
+def custom_gate_factory_from_pyquil_defgate(gate: pyquil.quil.DefGate):
+    num_qubits = int(math.log(gate.matrix.shape[0], 2))
```
simonwa7, line 245: If this is all `math` is used for, you can use `np.log2` as well I believe. Definitely not a necessary change however

src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
dexter2206: Those two arguments are defined as fixtures at the beginning of the file. Fixtures are part of pytest's dependency injection system. Basically, when you declare fixture with some name and then use the same name as a parameter, the pytest will call the fixture for you and populate the arguments. This usually reduces code in tests, but even more importantly, allows for controlling scope of such fixture. Currently, the fixtures in this module have a default "function" scope, which means they are called whenever some test function needs them.

src/python/zquantum/core/circuit/conversions/pyquil_conversions.py
dexter2206: Nice catch. I initially used numpy, but found out that it cannot handle arbitrarily large integers, which would put a limit on a circuit width. For instance `np.log2(2 ** 64)` already throws TypeError.

src/python/zquantum/core/circuit/conversions/pyquil_conversion_test.py
dexter2206: Thanks for catching this. Fixed in 8ab2891

PR #135, opened 2020-12-17
Fermion mode reordering

max-radin:
This PR adds a function and step for reordering the modes of interaction operators.


mstechly (approved):
Looks good to me 👍

PR #139, opened 2021-01-04
176036280 add testing that keep value history has effect for all optimizers

dexter2206:
This improves base optimizer tests:
- using fixtures/parametrization instead of loops
- testing if keep_value_history option is used correctly.
This requires refactoring tests of all optimizers, so this PR should be considered with corresponding PRs in z-quantum-optimizers and qe-qiskit.


src/python/zquantum/core/interfaces/optimizer_test.py
```diff
@@ -17,41 +18,83 @@ def sum_x_squared(x):
...
-            self.assertAlmostEqual(results.opt_params[0], 0, places=4)
-            self.assertAlmostEqual(results.opt_params[1], 0, places=4)
-
-            self.assertIn("nfev", results.keys())
-            self.assertIn("nit", results.keys())
-            self.assertIn("opt_value", results.keys())
-            self.assertIn("opt_params", results.keys())
-            self.assertIn("history", results.keys())
+    # To run tests with this class, Inherit it, and in the same module
+    # define a fixture with name optimizer that returns an optimizer.
```
mstechly, line 22: Maybe it would be a worthwhile to add an example of how to do that in the comment?

mstechly (approved):

src/python/zquantum/core/interfaces/optimizer_test.py
dexter2206: Maybe we should just add a reference to one of tests in optimizers? Wdyt?

mstechly: I think it's better to add a minimal example rather than reference an external repo.
Though on the other hand, they are all so connected that it's probably fine as well.
Whatever you think is best.

PR #143, opened 2021-01-06
Generate random list of integers

simonwa7:


mstechly (commented):
I don't see how it's "universally useful" enough to get into core?

mstechly (approved):

PR #145, opened 2021-01-07
176148013 backend gates tests

mstechly:


src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -117,6 +147,175 @@ def test_get_bitstring_distribution(self, backend):
...
+            # When
+            estimator = BasicEstimator()
+            expectation_value = estimator.get_estimated_expectation_values(
+                backend_for_gates_test,
+                circuit,
+                operator,
+            ).values[0]
+
+            # Then
+            assert expectation_value == pytest.approx(target_values[i], abs=sigma * 3)
```
simonwa7, line 190: This may be nit-picky, so feel free to ignore, but shouldn't we check the expectation values of the individual operators? Getting the sum of `I`, `X`, `Y`, and `Z` isn't as precise as measuring and comparing them separately

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -117,6 +147,175 @@ def test_get_bitstring_distribution(self, backend):
...
+        gate_2 = Gate(tested_gate, params=params, qubits=qubit_list)
+
+        circuit = Circuit()
+        circuit.gates = [gate_1, gate_2]
+        operators = [
+            QubitOperator("[]"),
+            QubitOperator("[X0]"),
+            QubitOperator("[Y0]"),
+            QubitOperator("[Z0]"),
+        ]
```
simonwa7, line 217: Same comment as above

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -175,3 +374,97 @@ def test_get_bitstring_distribution_wf_simulators(self, wf_simulator):
...
+        qubit_list = [Qubit(0)]
+        gate_1 = Gate(initial_gate, qubits=qubit_list)
+        gate_2 = Gate(tested_gate, qubits=qubit_list)
+
+        circuit = Circuit()
+        circuit.gates = [gate_1, gate_2]
+        import pdb
+
+        pdb.set_trace()
+
```
simonwa7, line 397: hello my old friend ;)

src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -175,3 +374,97 @@ def test_get_bitstring_distribution_wf_simulators(self, wf_simulator):
...
+    def test_two_qubit_parametric_gates_using_amplitudes(
+        self, wf_simulator, initial_gates, tested_gate, params, target_amplitudes
+    ):
+        # Given
+        qubit_list = [Qubit(0), Qubit(1)]
+        gate_1 = Gate(initial_gates[0], qubits=[qubit_list[0]])
+        if initial_gates[1] == "CNOT":
+            gate_2 = Gate(initial_gates[1], qubits=qubit_list)
+        else:
+            gate_2 = Gate(initial_gates[1], qubits=[qubit_list[1]])
```
simonwa7, line 460: Is only `initial_gates[1]` ever set to CNOT? No possibility for `initial_gates[0]`? Do we also make sure `len(initial_gates) == 2` always?

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
```diff
@@ -0,0 +1,333 @@
...
+)
+X = sympy.Matrix([[0, 1], [1, 0]])
+Y = sympy.Matrix([[0, -sympy.I], [sympy.I, 0]])
+Z = sympy.Matrix([[1, 0], [0, -1]])
+S = sympy.Matrix([[1, 0], [0, sympy.I]])
+T = sympy.Matrix([[1, 0], [0, (1 + sympy.I) / sympy.sqrt(2)]])
+
+CNOT = sympy.Matrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]])
+CZ = sympy.Matrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, -1]])
+SWAP = sympy.Matrix([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
```
simonwa7, line 20: This will be handled during rework, but just noting that this might cause "single source of truth" issues for gate definitions

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
```diff
@@ -0,0 +1,333 @@
...
+                )
+            exp_vals_string += "]"
+
+            print(
+                gate_names_string
+                + angle_string
+                + operator_names_string
+                + exp_vals_string
+                + "],"
+            )
```
simonwa7, line 206: It's hard to really know what's going on here. Maybe add some comments above each function saying what kinds of cases it is generating?

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
```diff
@@ -0,0 +1,333 @@
...
+    print("2 qubit gates")
+    print("-" * 10)
+    generate_cases_2_qubits_exp_vals_with_angles(CPHASE, "CPHASE", angles)
+    generate_cases_2_qubits_exp_vals_with_angles(XX, "XX", angles)
+    generate_cases_2_qubits_exp_vals_with_angles(YY, "YY", angles)
+    generate_cases_2_qubits_exp_vals_with_angles(ZZ, "ZZ", angles)
+
+
+if __name__ == "__main__":
+    main()
```
simonwa7, line 333: This seems like something that isn't supposed to be here anymore :)

src/python/zquantum/core/interfaces/backend_test.py
mstechly: We do check for every single operator. Check is happening inside a loop and we check for every single operator separately.

src/python/zquantum/core/interfaces/backend_test.py
mstechly: :)

src/python/zquantum/core/interfaces/backend_test.py
mstechly: Good catch, that's an artifact from some old code.

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
mstechly: This is helper code anyway, so I think we'll figure this out :)

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
mstechly: It actually should – these functions are not executed anywhere outside of this script, so leaving it here makes it easy to use and generate new cases :)

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
mstechly: I actually think that the names of the functions are pretty clear, do you have a suggestion what should I add to make it clearer?

src/python/zquantum/core/interfaces/backend_test.py
mstechly: Fixed in [bdab82](https://github.com/zapatacomputing/z-quantum-core/pull/145/commits/bdab8249a2b4f5b808b21b0ac7101368b29146a2).

src/python/zquantum/core/interfaces/backend_test.py
mstechly: Fixed in [bdab82](https://github.com/zapatacomputing/z-quantum-core/pull/145/commits/bdab8249a2b4f5b808b21b0ac7101368b29146a2)

src/python/zquantum/core/interfaces/backend_test.py
simonwa7: Whoops, my bad. Read this incorrectly :)

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
simonwa7: I was confused because I didn't understand why you were building strings and then printing them out without returning anything. When reading this, I actually initially thought it was creating some global variables or something to be used during testing. Now I realize these are those functions you used to generate the code for the different testing states. Maybe add a comment at the top of the file describing that this is what these functions are for and maybe when/why/how someone would alter them and/or use them?

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
simonwa7: Gotcha, this confusion goes along with this comment: https://github.com/zapatacomputing/z-quantum-core/pull/145#discussion_r554165870

Maybe a quick comment saying something like: `This script is used to generate test cases in the file <insert-filename>. Run with python3 <script>.py`

mstechly: Addressed in [46e52f](https://github.com/zapatacomputing/z-quantum-core/pull/145/commits/46e52f81e7ca383dc39a0b581a221a5a3c2693c3).

src/python/zquantum/core/testing/generate_cases_for_backend_tests.py
mstechly: Addressed in [46e52f](https://github.com/zapatacomputing/z-quantum-core/pull/145/commits/46e52f81e7ca383dc39a0b581a221a5a3c2693c3).

simonwa7 (approved):

PR #146, opened 2021-01-07
Moving grouping out of measurement estimation function, step 1

jgonthier:
The measurement estimation function can now accept a list of qubit operators, where each element in the list is a group of co-measurable terms, instead of accepting a raw qubit operator and a grouping method.

This is part of a refactoring effort to separate grouping from the measurement estimation.

The same work should be done at some point for the expectation value estimation functions.


src/python/zquantum/core/hamiltonian.py
```diff
@@ -159,9 +159,10 @@ def get_expectation_values_from_rdms(
 
 
 def estimate_nmeas(
-    target_operator: QubitOperator,
+    target_operator: Optional[QubitOperator] = None,
```
mstechly, line 162: I think having two different inputs is not a particularly good solution.
It would be better to have one input `target_operator_list` with type `Union[QubitOperator, List[QubitOperator]]`. It would accept a `QubitOperator` as a special case.
Much cleaner in my opinion, though still probably not ideal.

jgonthier: Ooh, you're right. I did not think about that. My main worry was to ensure backward compatibility with existing code.
Do you think it would be ok to just have `target_operator` with type `Union[QubitOperator, List[QubitOperator]]`, i.e. not change the name of the argument for now?

Eventually, it might make sense to completely get grouping out of the `estimate_nmeas` and expectation value estimation functions and then we'd probably have a cleaner function, but that will have to wait for later.

mstechly: I wouldn't worry about backward compatibility too much here, as it doesn't seem to be used in many places. We could just refactor these and it should be good. I like it more than having (even slightly) confusing variable names. But if you think we'll be refactoring soon I can live with that ;)
WDYT @simonwa7 @max-radin ?

simonwa7: Agreed with @mstechly, but I don't know how much of an issue this would be for backwards compatibility, so leaving this up to @jgonthier feels right to me.

max-radin: I feel like both `target_operator` and `target_operator_list` are going to be a bit confusing. I wonder if it would make sense to split this into two separate functions, one that acts on a list and one that acts on a QubitOperator. (Where the second would call the first.) This would solve the issue with variables name and give us slightly more modular code.

jgonthier: Sooo... `estimate_nmeas` and `estimate_nmeas_list`, @max-radin ? Sorry to be a bit slow to answer here guys.

max-radin: Ok I should be able to do that this afternoon!

max-radin (approved):

PR #147, opened 2021-01-11
Implemented calculation of correlations from measurements

max-radin:
This PR updates the `measurements.get_expectation_values` method to compute the correlations (expectation values of pairwise products of operator terms), which will be needed for estimating precisions of cost functions.

### DIFF
--- a/src/python/zquantum/core/measurement.py
+++ b/src/python/zquantum/core/measurement.py
@@ -114,7 +114,10 @@ class ExpectationValues:
     def to_dict(self) -> dict:
         """Convert to a dictionary"""
 
-        data = {"schema": SCHEMA_VERSION + "-expectation_values", "frames": []} # what is "frames" for?
+        data = {
+            "schema": SCHEMA_VERSION + "-expectation_values",
+            "frames": [],
+        }  # what is "frames" for?
 
         data["expectation_values"] = convert_array_to_dict(self.values)
 
@@ -148,6 +151,7 @@ class ExpectationValues:
                 covariances.append(convert_dict_to_array(covariance_matrix))
 
         return cls(expectation_values, correlations, covariances)
+
 
 def sample_from_wavefunction(
     wavefunction: Wavefunction, n_samples: int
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/measurement.py
```diff
@@ -500,7 +504,30 @@ def get_expectation_values(
                     value = -float(count) / num_measurements
                 expectation += np.real(coefficient) * value
             expectation_values.append(np.real(expectation))
-        return ExpectationValues(np.array(expectation_values))
+
+        correlations = np.zeros((len(ising_operator.terms),) * 2)
+        for i, first_term in enumerate(ising_operator.terms):
+            for j, second_term in enumerate(ising_operator.terms):
```
aakunitsa, line 510: I would like to suggest a few minor optimizations here:
- We could exclude diagonal correlations since those are squared Ising operator coefficients.
- There is no need to calculate the correlations of the constant term with the others; they can be efficiently extracted from the expectation values of the terms and corresponding coefficients.
- \<Ai Aj\> = \<Aj Ai\> so we could loop over (i,j) pairs such that e.g. i < j.

src/python/zquantum/core/measurement_test.py
```diff
@@ -586,12 +586,17 @@ def test_get_expectation_values_from_measurements(self):
         )
         ising_operator = IsingOperator("10[] + [Z0 Z1] - 10[Z1 Z2]")
```
aakunitsa, line 587: Maybe I am a bit paranoid here but would you mind making this test a bit less symmetric? I would add smth like 110 and 011 to the measurements and pick an Isining operator with other coefficients, e.g. 10, 2, -3. I remember a few cases when my buggy code passed symmetric tests and I spent countless hours hunting down those bugs later...

aakunitsa (changes requested):
Everything looks great and I am sure that the code is correct but I would like to request a few minor changes. Sorry :)

max-radin:
@aakunitsa I'm going to close this PR and address the comments in #148 instead.

src/python/zquantum/core/measurement_test.py
max-radin: Updated these tests in c1eabba0f84208cfdd63223a61b481a807c6ec2c

src/python/zquantum/core/measurement.py
max-radin: Updated this in the `precisions` branch

PR #148, opened 2021-01-12
Adding precision estimate to ansatz-based cost function

max-radin:


src/python/zquantum/core/measurement.py
```diff
@@ -500,7 +514,42 @@ def get_expectation_values(
                     value = -float(count) / num_measurements
                 expectation += np.real(coefficient) * value
             expectation_values.append(np.real(expectation))
-        return ExpectationValues(np.array(expectation_values))
+
+        correlations = np.zeros((len(ising_operator.terms),) * 2)
```
mstechly, line 518: It seems to me that you've added a lot of logic into this function and it's a little bit bloated – perhaps consider decomposing it into several functions to make it more legible?

src/python/zquantum/core/measurement.py
```diff
@@ -500,7 +514,42 @@ def get_expectation_values(
...
+        for i, first_term in enumerate(ising_operator.terms):
+            for j, second_term in enumerate(ising_operator.terms):
+                first_term_qubits = set(op[0] for op in first_term)
+                second_term_qubits = set(op[0] for op in second_term)
+                marked_qubits = first_term_qubits.symmetric_difference(
+                    second_term_qubits
+                )
+                for bitstring, count in bitstring_frequencies.items():
+                    bitstring_int = convert_bitstring_to_int(bitstring)
+                    if parity_even_p(bitstring_int, marked_qubits):
```
mstechly, line 528: Might be worth replacing `parity_even_p` from grove with our custom logic (see Slack)

aakunitsa: Also, I believe you are doing some unnecessary work when calculating correlations.  Diagonal elements of `correlations` are the squares of the `ising_operator` coefficients and the matrix itself is symmetric.

src/python/zquantum/core/measurement.py
```diff
@@ -500,7 +514,42 @@ def get_expectation_values(
...
+                            ising_operator.terms[first_term]
+                            * ising_operator.terms[second_term]
+                        )
+                        * value
+                    )
+
+        denominator = (
+            num_measurements - 1 if use_bessel_correction else num_measurements
+        )
+        covariances = np.zeros((len(ising_operator.terms),) * 2)
```
aakunitsa, line 543: `covariances = correlations - expectation_values[:, np.newaxis] *  expectation_values[np.newaxis, :]` should be faster and more compact.

src/python/zquantum/core/cost_function.py
```diff
@@ -5,11 +5,47 @@
...
+
+    value = np.sum(expectation_values.values)
+
+    precision = None
+
+    if expectation_values.covariances:
+        variance = 0
+        for frame_covariance in expectation_values.covariances:
+            variance += np.sum(frame_covariance, (0, 1))
+        precision = np.sqrt(variance)
```
aakunitsa, line 45: `precision` is a sqrt of the estimated variance of the ising operator but we actually need an estimate of the standard error (https://en.wikipedia.org/wiki/Standard_error). @jgonthier

aakunitsa (changes requested):
Sorry, I ended up requesting some changes for this PR.

src/python/zquantum/core/measurement.py
max-radin: Done!

src/python/zquantum/core/measurement.py
max-radin: I've modified the code to take advantage of symmetry and diagonal elements. @mstechly I'm going to put the replacement for `parity_even_p` in another PR because it gets used in places unrelated to this PR.

src/python/zquantum/core/measurement.py
max-radin: I've moved some duplicated logic into a new function `get_expectation_value_from_frequencies` and implemented @aakunitsa's suggestion to simplify the calculation of covariances.

src/python/zquantum/core/cost_function.py
max-radin: It actually is the standard error: `expectation_values.covariances` is the covariances of the estimators for the operator expectation values, not the covariances of the operators. I've renamed `covariances` to `estimator_covariances` and `variances` to `estimator_variance` to make this less confusing!

src/python/zquantum/core/measurement.py
mstechly: Sure, sounds good for me!

mstechly (approved):

aakunitsa (approved):
Thanks a lot for making all those changes @max-radin!

src/python/zquantum/core/cost_function.py
aakunitsa: Okay, gotcha. Sorry about that,  I got a bit confused!

src/python/zquantum/core/measurement.py
aakunitsa: Thanks!

PR #150, opened 2021-01-12
176466000 record backend usage

simonwa7:


src/python/zquantum/core/interfaces/backend.py
```diff
@@ -47,46 +53,64 @@ def run_circuitset_and_measure(
...
+            for circuit in circuit_set:
+                measurement_set.append(self.run_circuit_and_measure(circuit), **kwargs)
+            return measurement_set
+        else:
+            self.number_of_circuits_run += len(circuit_set)
+            self.number_of_jobs_run += 1
 
     def get_expectation_values(
-        self, circuit: Circuit, qubit_operator: QubitOperator, **kwargs
+        self, circuit: Circuit, operator: IsingOperator, **kwargs
```
mstechly, line 66: I know forest backend needs `IsingOperator`, but actually I think some others might work just well with `QubitOperator`.

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -114,7 +138,7 @@ def __init__(
         noise_model: Optional = None,
         device_connectivity: Optional[CircuitConnectivity] = None,
     ):
-        self.n_samples = n_samples
+        super().__init__(n_samples)
```
mstechly, line 141: I'm just curious what this change actually does :)

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -4,9 +4,10 @@
     create_bitstring_distribution_from_probability_distribution,
 )
 from ..circuit import Circuit, CircuitConnectivity
-from ..measurement import ExpectationValues, Measurements
+from ..measurement import ExpectationValues, Measurements, expectation_values_to_real
+from ..openfermion import expectation
```
mstechly, line 8: I think name of this function is a little bit too vague and not a verb.
`get_expectation_from_wavefunction` or sth like this?

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -128,25 +152,25 @@ def get_wavefunction(self, circuit: Circuit, **kwargs) -> Wavefunction:
...
             circuit (core.circuit.Circuit): quantum circuit to be executed.
-            qubit_operator(openfermion): Operator for which we calculate the expectation value.
+            operator(openfermion): Operator for which we calculate the expectation value.
 
         Returns:
             ExpectationValues: object representing expectation values for given operator.
         """
-
-        raise NotImplementedError
+        wavefunction = self.get_wavefunction(circuit)
```
mstechly, line 172: I'm not sure whether putting it like this actually gives us any benefits?
Some simulators might have special functions for expectation values (e.g. `qHipster`). 
It's not a lot of code for people to write it and having `raise NotImplementedError` is much clearer IMO.

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -128,25 +152,25 @@ def get_wavefunction(self, circuit: Circuit, **kwargs) -> Wavefunction:
             pyquil.Wafefunction: wavefunction object.
 
         """
-
-        raise NotImplementedError
+        self.number_of_circuits_run += 1
+        self.number_of_jobs_run += 1
```
mstechly, line 156: You've deleted `raise NotImplementedError`, now it will silently fail.
Also, it seems to me that perhaps it's worth adding this counter not inside the logic of the code, but rather as a decorator?

mstechly (changes requested):

src/python/zquantum/core/interfaces/backend.py
simonwa7: It lets us borrow the `init` functionality from `QuantumBackend` without rewriting the logic. Before, all QuantumBackend.__init__()` did was `self.n_samples = n_samples`, but now in addition, we set `self.number_of_circuits_run` and `self.number_of_jobs_run`.

src/python/zquantum/core/interfaces/backend.py
simonwa7: I agree, but I would suggest making a chore for this and addressing in a future PR since it will require unrelated  changes that need to be propogated

src/python/zquantum/core/interfaces/backend.py
simonwa7: Even though it's not a _lot_ of code, I still think it's good to have a default method that people can inherit without re-implementing. If someone wants to implement a new method, they can always override the default one

src/python/zquantum/core/interfaces/backend.py
simonwa7: I'm happy to reimplement this as a decorator but am actually wondering about when to use this style vs using a decorator.

src/python/zquantum/core/interfaces/backend.py
simonwa7: To clarify, the only reason we rewrite the `__init__` function at all is because we want to add `noise_model` and `device_connectivity` to the function header

mstechly: Sounds good!

src/python/zquantum/core/interfaces/backend.py
mstechly: Fine.

src/python/zquantum/core/interfaces/backend.py
mstechly: WDYT @dexter2206 ?

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -47,46 +57,65 @@ def run_circuitset_and_measure(
...
-        self, circuit: Circuit, qubit_operator: QubitOperator, **kwargs
+        self, circuit: Circuit, operator: SymbolicOperator, **kwargs
     ) -> ExpectationValues:
         """
         Executes the circuit and calculates the expectation values for given operator.
 
         Args:
             circuit (core.circuit.Circuit): quantum circuit to be executed.
-            qubit_operator(openfermion): Operator for which we calculate the expectation value.
+            operator(openfermion.IsingOperator): Operator for which we calculate the expectation value.
```
mstechly, line 77: You've changed the types in the function def, but not here :)

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -47,46 +57,65 @@ def run_circuitset_and_measure(
...
+        measurements = self.run_circuit_and_measure(circuit)
+        operator = change_operator_type(operator, IsingOperator)
+        expectation_values = measurements.get_expectation_values(operator)
+        expectation_values = expectation_values_to_real(expectation_values)
+        return expectation_values
 
     def get_expectation_values_for_circuitset(
-        self, circuitset: List[Circuit], qubit_operator: QubitOperator, **kwargs
-    ) -> [ExpectationValues]:
+        self, circuitset: List[Circuit], operator: IsingOperator, **kwargs
```
mstechly, line 89: You've changed it to `SymbolicOperator` in `get_expectation_values` but here it's `IsingOperator`. 
Please make it consistent.

mstechly (approved):

src/python/zquantum/core/interfaces/backend.py
simonwa7: Addressed here: 9f87e46

src/python/zquantum/core/interfaces/backend.py
simonwa7: Addressed here: 9f87e46

src/python/zquantum/core/interfaces/backend.py
dexter2206: I wouldn't recommend using a decorator here. `run_circuitset_and_measure` also increments those counters but by the number dependent on the input arguments and with some conditional logic, therefore it would be a bit tricky to write a single decorator that handles this. Even if we did that, the conditional logic of `run_circuitset_and_measure` would have to be repeated there. Or am I missing something?

mstechly: Sounds good if you say so 👍 
I'm in the phase where I would write all the code as decorators ;)

PR #151, opened 2021-01-13
176421730 update qiskit version

mstechly:


simonwa7 (commented):
I believe `src/setup.py` should be updated as well

simonwa7 (approved):

PR #154, opened 2021-01-13
Ground state cost function

simonwa7:


src/python/zquantum/core/cost_function_test.py
```diff
@@ -1,10 +1,84 @@
...
+from sympy import Symbol
+from .cost_function import AnsatzBasedCostFunction, get_ground_state_cost_function
 from .interfaces.mock_objects import MockQuantumSimulator, MockEstimator, MockAnsatz
 from openfermion import QubitOperator
 
+RNGSEED = 1234
+
+
+@pytest.fixture
+def ground_state_cost_function():
```
mstechly, line 13: I think it would be more scalable if you passed all (or at least some) the params as fixture params.

steps/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+from zquantum.core.utils import create_object
+from zquantum.core.serialization import save_optimization_results
+from openfermion import SymbolicOperator
+import json
+import os
+import numpy as np
+from typing import Union, Dict, Optional, List
+
+
+def optimize_parameterized_quantum_circuit(
```
mstechly, line 13: I think this name is slightly confusing – we're using `get_ground_state_cost_function` inside, and I believe that `optimize_parameterized_quantum_circuit` applies to a broader class of problems than just finding the ground state.

mstechly (changes requested):

src/python/zquantum/core/cost_function_test.py
simonwa7: Addressed here: 5c13656

steps/optimize.py
simonwa7: Addressed here: 
14a0ec7

simonwa7:
A quick question about `steps/optimize_test.py`. I definitely would like to keep these unittests and continue to write them in the future as it helps significantly with development and they could also become acceptance tests, however, I am worried that leaving them inside `steps` will bloat the directory. Does anyone have suggestions/thoughts on these kinds of files moving forward?

mstechly (approved):

PR #155, opened 2021-01-14
Added step for measurement analysis of grouped Hamiltonians

jgonthier:
There's also some Black formatting in the files I touched, and I added a utility function to get the Hartree-Fock RDM.


src/python/zquantum/core/utils.py
```diff
@@ -536,11 +539,39 @@ def load_nmeas_estimate(filename: str) -> Tuple[float, int, np.ndarray]:
...
-    K_coeff = data['K']
-    nterms = data['nterms']
+    frame_meas = convert_dict_to_array(data["frame_meas"])
+    K_coeff = data["K"]
+    nterms = data["nterms"]
 
     return K_coeff, nterms, frame_meas
+
+
+def hf_rdm(n_alpha: int, n_beta: int, n_orbitals: int):
```
max-radin, line 552: Can provide type hinting for the return value with `-> InteractionRDM`

max-radin (changes requested):
Could you add a unit test for `hf_rdm`? There might be one in zmachine. (Or just leave `hf_rdm` out of this PR.)

jgonthier:
I have some more modifications to make to this branch, actually, so I'll close the PR for now... Will open it back up once things are more settled.

src/python/zquantum/core/utils.py
jgonthier: Done now, and also added a test for `hf_rdm` (although it requires to store Hamiltonians which is quite ugly, if there's a better way please let me know)

jgonthier:
Aaaand reopening this PR for merging! I think it's big enough and any additional modifications I make should definitely be a separate PR.

steps/evaluation.py
```diff
@@ -51,6 +57,12 @@ def get_expectation_values_for_qubit_operator(
     save_expectation_values(expectation_values, "expectation-values.json")
 
 
+def get_ground_state_rdm_from_qubit_operator(qubit_operator, n_particles):
```
aakunitsa, line 60: Please add a docstring and specify the types of input parameters.

steps/measure.py
```diff
@@ -139,15 +152,53 @@ def hamiltonian_analysis(
...
+        nterms=nterms,
+        frame_meas=frame_meas,
+        filename="hamiltonian_analysis.json",
+    )
+
+
+def grouped_hamiltonian_analysis(
+    groups: str, expectation_values: str = "None",
+):
+
```
aakunitsa, line 169: Please add a docstring :)

steps/measure.py
```diff
@@ -139,15 +152,53 @@ def hamiltonian_analysis(
...
+    interactionrdm: str, qubit_operator: str, sort_terms: bool = False,
 ):
     operator = load_qubit_operator(qubit_operator)
     rdms = load_interaction_rdm(interactionrdm)
     expecval = get_expectation_values_from_rdms(rdms, operator, sort_terms=sort_terms)
-    save_expectation_values(expecval, 'expectation_values.json')
+    save_expectation_values(expecval, "expectation_values.json")
+
+
+def expectation_values_from_rdms_for_qubitoperator_list(
```
aakunitsa, line 196: Docstring pls!

aakunitsa (approved):
I haven't noticed any issues with this PR, but I think it would be good to add docstrings for the new steps and type annotations in some cases.

src/python/zquantum/core/hamiltonian.py
```diff
@@ -141,6 +145,36 @@ def compute_group_variances(
...
+    """
+
+    all_expectations = []
+    for qubitoperator in qubitoperator_list:
+        expectations = get_expectation_values_from_rdms(
+            interactionrdm, qubitoperator, sort_terms=sort_terms
+        )
+        all_expectations += list(expectations.values)
+
+    return ExpectationValues(np.asarray(all_expectations))
```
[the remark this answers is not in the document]
jgonthier: That's a good question! I thought about it but it would imply modifying other functions. I think we might have to change the ExpectationValues object later anyway to avoid the issues in https://www.pivotaltracker.com/story/show/176684130

When we take care of the above story we should think of a good way to return these ExpectationValues. My reasoning is also that the list of qubit operators is supposed to be representing the grouping of a single operator, hence the expectation values can live in a single object.

steps/evaluation.py
jgonthier: Addressed in 0ff7e3c086d6027c44ee79a7c917c44d3a7fdb48

steps/measure.py
jgonthier: Addressed in 0ff7e3c086d6027c44ee79a7c917c44d3a7fdb48

steps/measure.py
jgonthier: Addressed in 0ff7e3c086d6027c44ee79a7c917c44d3a7fdb48

src/python/zquantum/core/hamiltonian.py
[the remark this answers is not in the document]
aakunitsa: Thanks, Jerome! Makes sense!

src/python/zquantum/core/utils.py
max-radin: Best thing to do for large dicts is to put them in a separate file. This change is in aa99939d1922d0773f417c6815c88f402a1f2ddb

max-radin (approved):

PR #164, opened 2021-01-19
julia v1.5 dockerfile

brian-dellabetta:
Bumping Julia to v1.5 and fixing the symlink. Julia is now installed in /root rather than /app, in line with what z-quantum-default does.

FIRST ZAPATA PR


mstechly (approved):
Looks good to me 👍

simonwa7 (approved):

PR #165, opened 2021-01-20
Calculate expectation values properly now using mock backend

simonwa7:


testing/data_validation/get-expectation-values-for-qubit-operator.py
```diff
@@ -20,12 +20,19 @@
...
+        assert identity_term > 1.5
+        assert identity_term < 2.5
+
         for expectation_value in workflowresult[key]["expectation-values"][
             "expectation_values"
-        ]["real"]:
-            assert isinstance(expectation_value, float)
-            assert expectation_value > 0.0
-            assert expectation_value < 1.0
+        ]["real"][1:]:
```
mstechly, line 32: You're selecting a range here (`1:`) and then you ask whether it's `float`.
Shouldn't it be `[1]` ?

testing/data_validation/get-expectation-values-for-qubit-operator.py
```diff
@@ -20,12 +20,19 @@
             == "zapata-v1-expectation_values"
         )
 
+        identity_term = workflowresult[key]["expectation-values"]["expectation_values"][
+            "real"
+        ][0]
+        assert isinstance(float(identity_term), float)
+        assert identity_term > 1.5
```
mstechly, line 27: I can't see how this test actually works. No idea why the values are between 1.5 and 2.5. It looks like it loads data from some file, but it's done with command line, so I don't even know what file.

Could you add some comments to explain what file it's loading and make it clearer?

mstechly (changes requested):

testing/data_validation/get-expectation-values-for-qubit-operator.py
simonwa7: The data is loaded from the input file through the command line. So it gets run with `python3 data_validation_file.py workflowresult_file.json`. 

I would rather wait to update this and explain it better as a story later on since it's basically the entirety of the acceptance tests which we may redesign.

testing/data_validation/get-expectation-values-for-qubit-operator.py
simonwa7: No, we're for looping over the array and making sure each element is a float

mstechly: Oh, it's inside for!
Sorry, missed that!

testing/data_validation/get-expectation-values-for-qubit-operator.py
mstechly: Sounds good to me.

mstechly (approved):

PR #174, opened 2021-01-22
Run circuitset

akataba:
Added a v1 step to run a list of circuit and return a list of bitstring for each measurement


mstechly (approved):
Looks good!
Not sure about those indentations in the `.yaml` file, but if you say it's good, then I guess it is ;)

martamau (commented):
Maybe let's add a integration test/example workflow for this new step?

PR #181, opened 2021-01-28
Adding number of samples as parameter to QuantumBackend methods

max-radin:
This PR should not be merged until PRs adding support for number of samples to all the backends are ready.


src/python/zquantum/core/interfaces/backend_test.py
```diff
@@ -132,6 +152,33 @@ def test_run_circuitset_and_measure(self, backend):
         else:
             assert backend.number_of_jobs_run == number_of_circuits
 
+    def test_run_circuitset_and_measure_n_samples(self, backend):
+        # Note: this test may fail with noisy devices
```
mstechly, line 156: I think I know how to tackle this so that we can keep both clean structure and minimal code overview. Would be happy to talk about it.

mstechly (approved):

PR #184, opened 2021-01-28
Updated exponentiate function.

mstechly:


max-radin (changes requested):
There's two issues we have here: 

1. The new `exponentiate` function is equivalent to the `time_evolution` function with `trotter_order = 1`. (Aside from the support for different input types.)
2. The names and docstrings for `exponentiate` and `exponentiate_term` are a bit misleading because they are actually doing time evolution: e^(-iHt)

How about this: let's merge the new `exponentiate` into `time_evolution`, and rename `exponentiate_term` to something like `time_evolution_for_term` and update its docstring to clarify that it does time evolution. How does that sound?

max-radin (approved):

PR #187, opened 2021-02-01
176680489 add xy gate

dexter2206:
This adds XY gate implementation and respective conversions to and from cirq and pyquil.


src/python/zquantum/core/circuit/gates/_two_qubit_gates.py
```diff
@@ -98,3 +108,17 @@ def _create_matrix(self) -> sympy.Matrix:
                 [0, 0, 0, sympy.cos(arg) - 1j * sympy.sin(arg)],
             ]
         )
+
+
+class XY(TwoQubitRotationGate):
+    """Quantum XY gate."""
```
mstechly, line 114: So you have finally decided not to go with defining it as XX * YY?
Cause if you got with XX*YY, you'll get a different matrix, with `self.angle * 2`.

mstechly (approved):

PR #188, opened 2021-02-01
One-qubit hamiltonian

aakunitsa:
Implemented a step to extract 1-qubit Hamiltonians from `InteractionOperator`-s generated by run_psi4 (`qe-psi4`). 1-qubit Hamiltonians are defined by calculating the matrix elements (in second quantization, no fermion-to-qubit mapping is performed) <0011|H|0011>, <1100|H|1100>, <1100|H|0011>, and expanding the 2 x 2 matrix in the Pauli basis.


steps/operators.py
```diff
@@ -86,6 +89,48 @@ def reorder_fermionic_modes(
     reordered_operator = _reorder_fermionic_modes(interaction_operator, ordering)
     save_interaction_operator(reordered_operator, "reordered-operator.json")
 
+def get_one_qubit_hydrogen_hamiltonian(interaction_operator: Union[InteractionOperator, str]):
```
jgonthier, line 92: This is looking good! Could you add a test? Either in `steps/operators.py`, or moving the code below to a separate function and testing it there, I don't personally have a preference.

jgonthier (changes requested):
Just needs a little test!

steps/operators.py
aakunitsa: Good point! Addressed in 5c51763

steps/operators_test.py
```diff
@@ -45,3 +49,31 @@ def test_concatenate_qubit_operator_lists():
...
+    two_body_tensor = file_handle_2b.get('arr_0')
+    # Assemble interaction operator
+    h2_hamiltonian_int = InteractionOperator(constant=constant, one_body_tensor=one_body_tensor, two_body_tensor=two_body_tensor)
+    save_interaction_operator(h2_hamiltonian_int, 'interaction-operator.json')
+
+    get_one_qubit_hydrogen_hamiltonian('interaction-operator.json')
+    h2_1qubit = load_qubit_operator('qubit-operator.json')
+    h2_1qubit_sparse = get_sparse_operator(h2_1qubit, n_qubits=1)
+    h2_1qubit_dense = h2_1qubit_sparse.toarray()
+    print(h2_1qubit_dense)
```
jgonthier, line 70: Maybe we don't need that printing in the final test

steps/operators_test.py
```diff
@@ -45,3 +49,31 @@ def test_concatenate_qubit_operator_lists():
...
+
+    e_1q = eigvalsh(h2_1qubit_dense)
+
+    gs_4q = get_ground_state(get_sparse_operator(h2_hamiltonian_int))
+
+    assert isclose(e_1q[0], gs_4q[0])
+
+    os.remove('interaction-operator.json')
+    os.remove('qubit-operator.json')
```
jgonthier, line 79: I think we should remove files before the `assert`, to make sure we clean up before the test fails, for example.

steps/operators_test.py
```diff
@@ -45,3 +49,31 @@ def test_concatenate_qubit_operator_lists():
...
+    # Interaction Hamiltonian of H2 (0.74 A, STO-3G) generated using openfermion-psi4
+    constant = 0.7151043387432434
+    # Load one body tensor from .npz file
+    file_handle_1b = np.load('h2_one_body_tensor.npz')
+    one_body_tensor = file_handle_1b.get('arr_0')
+    # Load one body tensor from .npz file
+    file_handle_2b = np.load('h2_two_body_tensor.npz')
+    two_body_tensor = file_handle_2b.get('arr_0')
+    # Assemble interaction operator
+    h2_hamiltonian_int = InteractionOperator(constant=constant, one_body_tensor=one_body_tensor, two_body_tensor=two_body_tensor)
```
jgonthier, line 63: I guess you could also save the InteractionOperator from openfermion-psi4 directly as a json and use load_interaction_operator, but maybe it doesn't matter that much.

jgonthier (changes requested):
Just move the assert in the test after the code to remove files and it will be good!

steps/operators_test.py
aakunitsa: Good point! Will do!

steps/operators_test.py
aakunitsa: I have implemented all the changes in a5dd0fa. I thought CI might be failing because I relied on npz files to store one- and two-body coefficients so I just moved those tensors to the python file (if that's okay).

jgonthier (approved):
All good!

PR #191, opened 2021-02-03
176751833 circuit acceptance tests

simonwa7:
Note, this does incur a breaking change as the step `create_circuit_set_from_circuit_artifacts` was refactored into `batch_circuits`


testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,695 @@
...
+        filename = "params.json"
+        if os.path.exists(filename):
+            os.remove(filename)
+
+        # When
+        generate_random_ansatz_params(ansatz_specs=ansatz_specs, seed=seed)
+
+        # Then
+        assert os.path.exists(filename)
+        parameters = load_circuit_template_params(filename)
```
mstechly, line 69: The thing we're doing here is ridiculous.
Steps really should be python functions that return python objects.
And these python objects should be serialized in the runtime (or something).
Sorry...
It's not the place nor time for such discussion...

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,695 @@
...
+        assert len(parameters) == number_of_layers
+        for parameter in parameters:
+            assert parameter < np.pi * 0.5
```
mstechly, line 72: I think these assertions should be removed.
If we ever change the default values of the parameters, these tests will start to fail.
If think it would be better to pass some `min_value` and `max_value` as input param and then check if parameters are set correctly.
And there could be a separate test which checks whether default values are `-pi/2;+pi/2`. WDYT?

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,695 @@
...
+            (1, 2, 2, 0, 1, 0.5),
+            (1, 0, 2, 0, 1, 0.5),
+            (6, 2, 2, 0, 1, 0.5),
+            (1, 2, 6, 0, 1, 0.5),
+            (1, 2, 6, -np.pi, 1, 0.5),
+            (1, 2, 6, -np.pi, np.pi, 0.5),
+            (1, 2, 6, 0, 1, 0.01),
+        ],
+    )
+    def test_build_uniform_param_grid_ansatz_specs_as_string(
```
mstechly, line 293: Why we do have this test here, but we didn't have similar test for `Test_build_ansatz_circuit` ?

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,695 @@
...
+        circuit_layers = load_circuit_layers(expected_circuit_layers_filename)
+        circuit_connectivity = load_circuit_connectivity(
+            expected_circuit_connectivity_filename
+        )
+
+        os.remove(expected_circuit_layers_filename)
+        os.remove(expected_circuit_connectivity_filename)
+
+
+class Test_create_random_circuit:
```
mstechly, line 426: All the class names should be CamelCase, right?

mstechly (changes requested):

steps/circuit.py
```diff
@@ -18,28 +18,31 @@
...
 from zquantum.core.utils import create_object
 from zquantum.core.testing import create_random_circuit as _create_random_circuit
-from typing import Dict, Union
+from typing import Dict, Union, List
 
 # Generate random parameters for an ansatz
 def generate_random_ansatz_params(
-    ansatz_specs: Dict,
-    number_of_parameters: Union[str, int] = "None",
+    ansatz_specs: Union[str, Dict] = None,
```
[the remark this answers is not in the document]
simonwa7: Yep :) this might break some workflows, but it's a very simple fix

testing/acceptance_tests/circuit_test.py
simonwa7: Agreed - I might take some time tomorrow for a small innovation day to see if we can't add something simple to the runtime to handle this

simonwa7: Maybe the first step would be something simple where you can return basic types and the runtime will serialize it for you. Then we can build off of that to add support for our types. Because all we need is a way to make a map from type -> serialization method. The question just then becomes how to make that map update when you have different resources installed

testing/acceptance_tests/circuit_test.py
simonwa7: After thinking on it more, I think it just makes sense to remove the assertions that the parameters are within the given range. I feel like that's something we can _trust_ since it's really coming from numpy. Thoughts?

steps/circuit.py
```diff
@@ -92,10 +102,9 @@ def build_uniform_param_grid(
 # Build circuit layers and connectivity
 def build_circuit_layers_and_connectivity(
     x_dimension: int,
-    y_dimension: Union[int, str] = "None",
+    y_dimension: int = None,
```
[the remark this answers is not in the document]
simonwa7: Addressed here: 64f6e67

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,695 @@
...
+    @pytest.mark.parametrize(
+        "number_of_layers, problem_size",
+        [
+            (1, 1),
+            (0, 1),
+            (1, 6),
+            (4, 3),
+        ],
+    )
+    def test_generate_random_ansatz_params_using_mock_ansatz_specs(
```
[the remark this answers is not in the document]
simonwa7: Addressed here: 9e6a714

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed here: 9e6a714

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed here: 9e6a714

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed here: 
f31c6d8

steps/circuit.py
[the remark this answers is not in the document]
mstechly: Then please make sure that you update all the example workflows and announce this as a warning to all the users :)

mstechly:
Seems good to me, but:
a) there are still a couple of other comments to address
b) will wait for other people to comment :)

steps/circuit.py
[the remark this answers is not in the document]
simonwa7: I'm wondering if perhaps it makes sense to wait to update the example workflows until after the runtime type-checking is updated. Since that should be coming soon it wouldn't make sense to update them twice.

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,744 @@
...
+            os.remove(filename)
+
+        # When
+        generate_random_ansatz_params(
+            number_of_parameters=number_of_parameters, seed=seed
+        )
+
+        # Then
+        assert os.path.exists(filename)
+        parameters = load_circuit_template_params(filename)
```
dexter2206, line 118: Technically this assert is not needed here, as the test will fail on `load_circuit_template_params` if file `filename` does not exist or is not readable.

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,744 @@
...
+        number_of_parameters = 2
+        ansatz_specs = {
+            "module_name": "zquantum.core.interfaces.mock_objects",
+            "function_name": "MockAnsatz",
+            "number_of_layers": 2,
+            "problem_size": 1,
+        }
+        seed = RNDSEED
+
+        # When
```
dexter2206, line 137: This is an example where annotating with given/when/then doesn't quite fit, right? How about we delete this comments from this and other tests that don't have "then"? Otherwise we can mark line with `raises` with "then" and line with call to `generate_random_ansatz_params` with "when".

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,744 @@
...
+            "problem_size": 2,
+        }
+
+        # When
+        build_ansatz_circuit(ansatz_specs=json.dumps(ansatz_specs))
+
+        # Then
+        circuit_filename = "circuit.json"
+        circuit = load_circuit(circuit_filename)
+        assert isinstance(circuit, Circuit)
```
dexter2206, line 261: In ideal world this would be already verified by a static type checker. I wonder if there are other, stronger assertions we might want to verify about the circuit we just loaded?

dexter2206 (changes requested):
My main concerns are about handling IO and possibility of not cleaning up after tests. I also lest some minor improvement suggestions.

steps/circuit.py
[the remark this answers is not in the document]
mstechly: Fine, resolving.

steps/circuit.py
```diff
@@ -105,46 +114,47 @@ def build_circuit_layers_and_connectivity(
...
 
-# Create circuitset from circuit artifacts
-def create_circuit_set_from_circuit_artifacts(
-    circuit1: str,
-    circuit2: str = "None",
-    circuit3: str = "None",
-    circuit4: str = "None",
-    circuit_set: str = "None",
+# Create one circuitset from circuit and circuitset objects
+def batch_circuits(
```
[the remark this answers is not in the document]
mstechly: I think we had this discussion elsewhere, resolving.

testing/acceptance_tests/optimize_test.py
```diff
@@ -137,3 +131,4 @@ def test_optimize_parametrized_circuit_for_ground_state_of_operator_optimizer_sp
     os.remove("optimization_results.json")
     os.remove("circuit.json")
     os.remove("initial_parameters.json")
+    os.remove("fixed_parameters.json")
```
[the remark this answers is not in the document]
simonwa7: This should be addressed here: https://www.pivotaltracker.com/story/show/176896194

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed here: 0f15633

testing/acceptance_tests/circuit_test.py
simonwa7: Addressed

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,744 @@
...
+            ansatz_specs=ansatz_specs,
+            number_of_layers=number_of_layers,
+            min_value=min_value,
+            max_value=max_value,
+            step=step,
+        )
+
+        # Then
+        param_grid = load_parameter_grid(expected_parameter_grid_filename)
+        os.remove(expected_parameter_grid_filename)
```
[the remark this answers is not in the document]
simonwa7: Addressed

testing/acceptance_tests/circuit_test.py
```diff
@@ -0,0 +1,744 @@
...
+        ansatz_specs = {
+            "module_name": "zquantum.core.interfaces.mock_objects",
+            "function_name": "MockAnsatz",
+            "number_of_layers": number_of_layers,
+            "problem_size": 2,
+        }
+        seed = RNDSEED
+
+        filename = "params.json"
+        if os.path.exists(filename):
```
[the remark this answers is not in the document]
simonwa7: Addressed

steps/circuit.py
```diff
@@ -70,17 +75,22 @@ def build_ansatz_circuit(ansatz_specs: Dict, params: str = "None"):
 
 # Build uniform parameter grid
 def build_uniform_param_grid(
-    ansatz_specs: Dict,
-    number_of_params_per_layer: Union[str, int] = "None",
+    ansatz_specs: Optional[Union[str, Dict]] = None,
```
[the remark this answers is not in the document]
simonwa7: Addressed here: d1549e4

steps/circuit.py
```diff
@@ -70,17 +75,22 @@ def build_ansatz_circuit(ansatz_specs: Dict, params: str = "None"):
...
 ):
-    if ansatz_specs != "None":  # TODO None issue in workflow v1
-        ansatz = create_object(json.loads(ansatz_specs))
+    assert (ansatz_specs is None) or (number_of_params_per_layer is None)
+    assert not ((ansatz_specs is None) and (number_of_params_per_layer is None))
+
+    if ansatz_specs is not None:
+        if isinstance(ansatz_specs, str):
+            ansatz_specs = json.loads(ansatz_specs)
+        ansatz = create_object(ansatz_specs)
```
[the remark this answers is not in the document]
simonwa7: Addressed here: 8321aad

steps/circuit.py
```diff
@@ -18,28 +18,31 @@
...
+    ansatz_specs: Optional[Union[str, Dict]] = None,
+    number_of_parameters: Optional[int] = None,
     min_value: float = -np.pi * 0.5,
     max_value: float = np.pi * 0.5,
-    seed: Union[str, int] = "None",
+    seed: Optional[int] = None,
 ):
-    if ansatz_specs != "None":  # TODO None issue in workflow v1
+    assert (ansatz_specs is None) or (number_of_parameters is None)
+    assert not ((ansatz_specs is None) and (number_of_parameters is None))
```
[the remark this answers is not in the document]
simonwa7: Addressed here: 8321aad

mstechly (approved):
Waiting for @dexter2206 to review the changes, but for me it looks good.

dexter2206 (approved):

PR #193, opened 2021-02-04
Adding optimal shot allocation to BasicEstimator

max-radin:
This allows the BasicEstimator to perform optimal shot assignment, with or without using prior expectation values to estimate variances.

Note that because this adds a new optional input to the `evaluate_ansatz_based_cost_function` step, it will break backwards compatibility with workflows using it.


mstechly:
@max-radin Is shot allocation something that could be abstracted from the estimator? Is there a case, where you would like to use the same method of shot allocation with a different optimizer?

src/python/zquantum/core/estimator.py
```diff
@@ -102,23 +106,42 @@ def get_estimated_expectation_values(
...
-            delta (float): a confidence term.
+            backend: the backend that will be used to run the circuit
+            circuit: the circuit that prepares the state.
+            target_operator): List of target functions to be estimated.
+            n_samples: Number of measurements to be performed on each frame.
+                Exactly one of n_samples and n_total_samples must be provided.
+            n_total_samples: Total number of measurements to be performed across
+                all frames. Exactly one of n_samples and n_total_samples must be
+                provided.
+            epsilon: Inherited from Estimator, not used.
```
mstechly, line 126: I think that if `epsilon` and `delta` are just `Inherited from Estimator, not used` – we should change the `Estimator` interface so that they're are not required.
I had issues with these elsewhere too and they don't think like a good part of the `Estimator` interface.

src/python/zquantum/core/estimator.py
```diff
@@ -129,17 +152,46 @@ def get_estimated_expectation_values(
...
+                        Ignoring backend.n_samples={backend.n_samples}"""
+                )
+                n_samples = (n_samples,) * len(frame_circuits)
+                measurements_set = backend.run_circuitset_and_measure(
+                    frame_circuits, n_samples
+                )
+            else:
+                measurements_set = backend.run_circuitset_and_measure(frame_circuits)
+
+        elif shot_allocation_strategy == "optimal":
```
mstechly, line 174: It looks to me that since we have two very different cases, with two different input params, maybe these should be two different classes?

src/python/zquantum/core/estimator.py
```diff
@@ -129,17 +152,46 @@ def get_estimated_expectation_values(
...
+                raise ValueError(
+                    "For optimal shot allocation, n_total_samples must be provided."
+                )
+
+            if n_samples is not None:
+                raise ValueError(
+                    "Optimal shot allocation does not support n_samples; use n_total_samples instead."
+                )
+
+            K2, nterms, measurements_per_frame = estimate_nmeas_for_frames(
```
mstechly, line 185: What's the `K2` ? It doesn't seem to be used anywhere?

src/python/zquantum/core/estimator_test.py
```diff
@@ -163,6 +164,79 @@ def test_get_estimated_expectation_values_with_constant(
         assert len(values) == 2
         assert coefficient == value
 
+    def test_get_estimated_expectation_values_optimal_shot_allocation(
+        self, estimator, backend, circuit, operator, epsilon, delta
+    ):
+        # Given
```
mstechly, line 170: I think this test doesn't really check whether the shot allocation strategy is working correctly. That's probably due to the use of `MockBackend` – I would add a `TODO` here, or a story to the backlog, to add more robust tests once the we replace `MockBackend` with `SimpleBackend`.

mstechly (changes requested):

src/python/zquantum/core/estimator.py
max-radin: I think that `epsilon` (desired precision) might make sense to keep, although it could use a more descriptive name. We could make `BasicEstimator` support it, just haven't gotten around to it yet. I'm less certain that we should keep `delta`, which I think @sabeegrewal originally intended to use for specifying the desired bound on the error according to some confidence level.

src/python/zquantum/core/estimator_test.py
max-radin: Yup, added a todo for the two optimal shot allocation tests in 1913685791dc9c11d72909115a85591e9c9d51a6.

src/python/zquantum/core/estimator.py
max-radin: It's the proportionality constant between number of measurements and inverse square precision; see [docstring](https://github.com/zapatacomputing/z-quantum-core/blob/f9a3de0c9a7cd4c654d928a03ad52062cebf3dce/src/python/zquantum/core/hamiltonian.py#L275). Anyway I replaced with `_` since it's not used here.

src/python/zquantum/core/estimator.py
max-radin: I combined some of the duplicated code across the different cases in cf9fdb6245461d9cfaca77dfc879a55905e9c315. If/when we add support for `n_total_samples` to the `shot_allocation_strategy="uniform"` case, then the two cases will be more similar.

Splitting it into two classes though wouldn't address the remaining edge case (that `n_samples` can't be used with `shot_allocation_strategy="optimal"`) because you'd still inherit `n_samples` from the base `Estimator` class. Perhaps we should change `Estimator` to have `n_total_samples` instead of `n_sampels`? Although this feels like a story for another day.

max-radin:
@mstechly regarding abstracting shot allocation from the estimator... interesting question! It's not yet clear to me how far up we can abstract shot allocation because the different estimators are allocating shots amongst different things. For example, while the basic estimator allocates shots over groups of co-measurable Hamiltonian terms, the enhanced sampling estimator involves allocating shots over unitaries and number of Grover iterates. So I think for now it would make sense to implement allocation separately for different estimators.

src/python/zquantum/core/estimator.py
mstechly: Ok, I'll defer to your judgment here, as I don't have a good feeling about what would be best.
Also, it might be reasonable to deal with that in a separate PR.

src/python/zquantum/core/estimator.py
mstechly: 👍

mstechly (approved):
Ok, looks good now!
Just please add relevant stories to the backlog :)

PR #198, opened 2021-02-10
176793572 determining exact bitstring frequencies

simonwa7:


src/python/zquantum/core/measurement.py
```diff
@@ -446,6 +447,36 @@ def from_counts(cls, counts: Dict[str, int]):
...
+
+        bitstring_samples = []
+        for state in distribution:
+            bitstring = tuple([int(measurement_value) for measurement_value in state])
+
+            bitstring_samples += [bitstring] * int(
+                distribution[state] * number_of_samples
+            )
+
+        while len(bitstring_samples) != number_of_samples:
```
max-radin, line 464: Can we delete the while loop since the needed bitstrings will get added during the first iteration?

src/python/zquantum/core/measurement.py
```diff
@@ -446,6 +447,36 @@ def from_counts(cls, counts: Dict[str, int]):
         measurements.add_counts(counts)
         return measurements
 
+    @classmethod
+    def get_measurements_representing_distribution(
```
max-radin, line 451: Could you add type hints and a brief docstring?

src/python/zquantum/core/measurement_test.py
```diff
@@ -31,223 +30,271 @@
...
+    assert np.allclose(
+        expectation_values.estimator_covariances[1], np.array([[0.00428797]])
+    )
+
+    assert np.allclose(
+        expectation_values.estimator_covariances[2], np.array([[0.0075706]])
+    )
+
+
+def test_expectation_vafmneaslues_to_real():
```
max-radin, line 196: Looks like this test name got mangled, should be `test_expectation_values_to_real `

max-radin (changes requested):
For the two tests `test_get_measurements_representing_distribution_correctly_samples_leftover_bitstrings` and `test_get_measurements_representing_distribution_randomly_samples_leftover_bitstrings_when_probabilities_equal`, I think it would be good to set a seed for the random number generator so that we can avoid a situation down the road where the tests randomly fail. (In particular it's not obvious how frequently the second test would fail.)

Looking at the [lea docs](https://bitbucket.org/piedenis/lea/wiki/Lea3_Tutorial_1#markdown-header-random-sampling-random-random_draw-methods), it looks like all we need to do is call `random.seed(0)` at the beginning of the test.

src/python/zquantum/core/measurement_test.py
simonwa7: Whoops... I'm gonna blame that one on Summit 😆

simonwa7:
I set the values so that the chance of a random failure was << 0.01 %, but setting the seed is better

src/python/zquantum/core/measurement.py
simonwa7: Addressed here: 2fb7d11

src/python/zquantum/core/measurement.py
simonwa7: Addressed here: e9b5689

src/python/zquantum/core/measurement_test.py
simonwa7: Addressed here: 
cedc8bb

max-radin (approved):

PR #199, opened 2021-02-10
Renamed expectation function to get_expectation_value

mstechly:


src/python/zquantum/core/openfermion/_utils.py
```diff
@@ -361,7 +361,7 @@ def reverse_qubit_order(qubit_operator: QubitOperator, n_qubits: Optional[int] =
     return reversed_op
 
 
-def expectation(qubit_op, wavefunction, reverse_operator=True):
+def get_expectation_value(qubit_op, wavefunction, reverse_operator=True):
```
alexjuda, line 364: Does this break the public API?

alexjuda: If you really dislike the old name you can add another function that just wraps the old one, this way it would be backwards-compatible

mstechly: It's not used in that many places, so I just fixed all the occurrences inside `z-quantum-core` and checked it's not being explicitly used outside `z-quantum-core`.
Also, since we're still in 0.x.x versions, I think this kind of change shouldn't be too surprising?

alexjuda (approved):

PR #197, opened 2021-02-10
4585620 WIP package with reworked circuits and gates

dexter2206:
This PR introduces new `zquantum.core.wip` package containing reworked circuits, gates and conversion functions. Changes should be fully backwards compatible (i.e. using of this new package is entirely optional).

Further plans for this package are as follows:
- Reworked circuits and gates stay in `wip` package until rework is fully completed
- Once the rework is finished, old circuits will be marked as deprecated
- After some transition period, old circuits/gates implementation will be removed and replaced by the reworked one.

PR #204, opened 2021-02-17
Qiskit circuit translations

alexjuda:
Adds support for translating whole circuits between Orquestra and Qiskit. Builds upon changes from #197 where only gate-level translation was supported. Developed in pair with @dexter2206.

Additionally, the PR contains a subtle refactor of the Circuit class:
- adds `n_qubits` property to allow representing circuits with unused qubits. Qiskit allows it already, so it was necessary for converting between the two frameworks.
- removes the `qubits` property, as I couldn't think of use cases for it and it provided no additional information over the afromentioned `n_qubits`. 
- removes support for passing `serializable=False` flag to `to_dict()` method. Like above, I couldn't think of a use-case for generating non-serializable dictionaries. 
- reimplements the `__add__` operator, so it supports concatenating two circuits together horizontally + to allow defining the circuit with gates in a similar way to PyQuil, while keeping the objects immutable, like:
```
circuit = Circuit()
circuit += X(0)
circuit += RX(0, 0)
```

Considering the code removals – I believe we should go for a codebase that's as small as possible, with **just right** feature  set size. Smaller code usually means easier software design and less maintenance + there's no point in having a feature if nobody uses it.

If you find a use-case that I didn't, please let me know! I'm open to re-thinking how to address that need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ dmypy.json
 .pyre/
 
 # End of https://www.gitignore.io/api/python
+
+# virtual env kept in project dir
+venv
--- a/src/python/zquantum/core/wip/circuit/_circuit.py
+++ b/src/python/zquantum/core/wip/circuit/_circuit.py
@@ -1,9 +1,9 @@
 import json
-from typing import List, Dict, Union, TextIO, Iterable, Any
+from typing import Dict, Union, TextIO, Iterable
+from functools import reduce
+
 from .gates import Gate
 from ...utils import SCHEMA_VERSION
-
-import sympy
 
 
 class Circuit:
@@ -27,17 +27,8 @@ class Circuit:
 
     @property
     def symbolic_params(self):
-        """ The set of symbolic parameters used in the circuit
-
-        Returns:
-            set: set of all the sympy symbols used as params of gates in the circuit.
-        """
-        symbolic_params = []
-        for gate in self.gates:
-            symbolic_params_per_gate = gate.symbolic_params
-            symbolic_params += symbolic_params_per_gate
-
-        return set(symbolic_params)
+        """Set of all the sympy symbols used as params of gates in the circuit."""
+        return reduce(set.union, (set(gate.symbolic_params) for gate in self._gates))
 
     def __eq__(self, other: "Circuit"):
         if not isinstance(other, type(self)):
@@ -58,8 +49,8 @@ class Circuit:
         new_circuit.gates = self.gates + other_circuit.gates
         return new_circuit
 
-    def evaluate(self, symbols_map: Dict["sympy.Symbol", Any]):
-        """Create a copy of the current Circuit with the parameters of each gate evaluated to the values
+    def evaluate(self, symbols_map: Dict):
+        """ Create a copy of the current Circuit with the parameters of each gate evaluated to the values 
         provided in the input symbols map
 
         Args:
--- a/src/python/zquantum/core/wip/circuit/_circuit.py
+++ b/src/python/zquantum/core/wip/circuit/_circuit.py
@@ -113,3 +113,6 @@ class Circuit:
 
         gates = [Gate.load(gate_data) for gate_data in data["gates"]]
         return cls(gates=gates)
+
+    def __repr__(self):
+        return f"{type(self).__name__}(gates={self.gates}, n_qubits={self.n_qubits})"
--- a/src/python/zquantum/core/wip/circuit/_circuit.py
+++ b/src/python/zquantum/core/wip/circuit/_circuit.py
@@ -1,46 +1,29 @@
 import json
+from typing import List, Dict, Union, TextIO, Iterable, Any
 from .gates import Gate
 from ...utils import SCHEMA_VERSION
-from typing import List, Dict, Union, TextIO, Any
 
 import sympy
 
 
-class Circuit(object):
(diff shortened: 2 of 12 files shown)


alexjuda:
Woah, we've squashed the base branch before merging to dev, so I gotta do some history rewriting!

alexjuda:
> Woah, we've squashed the base branch before merging to dev, so I gotta do some history rewriting!

done

.gitignore
```diff
@@ -108,3 +108,6 @@ dmypy.json
 .pyre/
 
 # End of https://www.gitignore.io/api/python
+
+# virtual env kept in project dir
+venv
```
mstechly, line 113: Since I call all my venvs differently, they just all include the string `venv`, perhaps going with a regex would be better?
Not sure though.

src/python/zquantum/core/wip/circuit/_circuit.py
```diff
@@ -95,44 +72,40 @@ def evaluate(self, symbols_map: Dict["sympy.Symbol", Any]):
         evaluated_circuit = circuit_class(gates=evaluated_gate_list)
         return evaluated_circuit
 
-    def to_dict(self, serializable: bool = True):
-        """ Creates a dictionary representing a circuit.
+    def to_dict(self):
```
mstechly, line 75: What was the main motivation behind using `serializable` arg?
@max-radin @simonwa7 ?
Was it the fact that an issue with numpy arrays, sympy symbols or sth else?

src/python/zquantum/core/wip/circuit/_circuit_test.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import numpy.random
```
mstechly, line 2: I think this import is unnecessary.

src/python/zquantum/core/wip/circuit/conversions/qiskit_conversions.py
```diff
@@ -52,11 +55,35 @@
...
+# Qiskit implements a special subclasses for some of the controlled gate variants:
+# https://github.com/Qiskit/qiskit-terra/blob/0.16.4/qiskit/circuit/add_control.py#L72
+# We'll support most of them eventually. The development plan is:
+# 1. Add mapping for some most common cases (done).
+# 2. Support most of the Qiskit controlled gates via our CustomGate when it's reworked.
+# 3. Add specialized gate subclasses on our side if there's a need for handling special
+#     cases explicitly.
+
+
 QISKIT_TO_ORQUESTRA_MAPPING = {
```
mstechly, line 79: I think we don't use "ORQUESTRA" anywhere in the code really, rather "zquantum".
TBH I'm not sure whether that's a good or bad decision, just pointing this out.
CC: @max-radin

alexjuda: The `ORQUESTRA` word here follows a convention used previously in other reworked conversion modules (`cirq_conversions`, `pyquil_conversions`), as well as some other functions and docstrings under the `wip` package. TBH I like `zquantum` better instead of `orquestra` in the code here because it would be consistent with both the repo and the package name. If we agree about it, I'd do this in another PR. WDYT, @dexter2206?

src/python/zquantum/core/wip/circuit/_circuit_test.py
alexjuda: Right. (2735047)

src/python/zquantum/core/wip/circuit/_circuit.py
simonwa7: I was recently chatting with @alexjuda about this and I couldn't really remember why we chose to use it in the first place. I can't think of any time you would want a `dict` representation of a `Circuit` or a `Gate` without wanting it to be serializable.

.gitignore
mstechly: I keep them outside of the projects directories, so they have a clear separation :)
Ok, I get the regex argumentation, looks good to me 👍

src/python/zquantum/core/wip/circuit/_circuit.py
mstechly: Sounds good, thanks @simonwa7  👍

src/python/zquantum/core/wip/circuit/conversions/qiskit_conversions.py
mstechly: I definitely won't mind that, but still would like to learn from @max-radin whether there are some other good reasons for not including the keyword "orquestra" in our python codebase :)

max-radin: At some point in the past we decided to drop `orquestra` from our python libraries so that people wouldn't mix up libraries that can be run on the platform with the underlying platform itself. Maybe someday we'll rename things, but for now I think it would be good to use `zquantum` here for consistency.

alexjuda: I've done the rename for functions and docstrings under the `wip` package (38c22f4). It was quick and we don't risk much by doing it already:
- `ORQUESTRA` -> `ZQUANTUM`
- `Orquestra` -> `ZQuantum`
- `orquestra` -> `zquantum`

mstechly (approved):

PR #205, opened 2021-02-17
Refactored estimator interface.

mstechly:
Given that I switched from specific params to `**kwargs`, this doesn't break other modules :)


dexter2206 (approved):
Looks so much better than previously. I recommend considering passing `estimator_kwargs` as  actual kwargs (so the signatures end in `**estimator_kwargs`) - this would simplify checking if it is `None`, as keyword args dictionary is never `None`.

mstechly:
@dexter2206 I've refactored it, but actually, now I see that it wasn't exactly what you suggested :D 
I don't want to pass `estimator_kwargs` in places like `get_ground_state_cost_function`, as there might be other kwargs for other objects.
Let me know if the changes I did make sense at all.

PR #207, opened 2021-02-18
Removing __init__.py from zquantum

max-radin:
Having an `__init__.py` in `src/python/zquantum` prevents the importing of other packages in the `zquantum` namespace such as `z-quantum-inference`.


mstechly (approved):
I just hope it doesn't break anything :)

max-radin:
@dexter2206 it seems that the unit tests for bitstring distribution do not work when the `__init__.py` is removed from `src/python/zquantum`. However having this file causes other packages in the `zquantum` namespace not to work. What do you think is a good way to address this?

dexter2206:
@max-radin fixed in 858880c :)

PR #208, opened 2021-02-18
Typo fix in step

max-radin:
Also includes bugfix from #203


mstechly (commented):
Why have you added `load_from_specs` here?
nvmd now I see more context, looking into it again.

mstechly (approved):

PR #210, opened 2021-02-19
Allowing circuits and qubit operators to be dict inputs

max-radin:


steps/evaluation.py
```diff
@@ -35,13 +35,14 @@
...
     evaluate_qubit_operator_list as _evaluate_qubit_operator_list,
+    convert_dict_to_qubitop,
 )
 
 
 def get_expectation_values_for_qubit_operator(
     backend_specs: Union[Dict, str],
-    circuit: Union[str, Circuit],
-    qubit_operator: Union[str, SymbolicOperator],
+    circuit: Union[str, Circuit, Dict],
```
mstechly, line 44: I think that the whole serialization/deserialization logic should be inside steps and not here.
The only type that we should allow for should be `Circuit`.
Though in reality this will require a bigger refactor anyway so I think it's fine for now.

mstechly (approved):

mstechly:
@max-radin any updates here?

steps/evaluation.py
mstechly: Damn... It is a step already 🤔 
I think I got confused somewhere and this comment is not valid.

PR #211, opened 2021-02-22
177008028 refactor estimator tests

mstechly:


src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -24,7 +24,10 @@ def __init__(self, options: Optional[Dict] = None):
 
     @abstractmethod
     def minimize(
-        self, cost_function: CallableWithGradient, initial_params: np.ndarray, **kwargs
+        self,
+        cost_function: Union[CallableWithGradient, callable],
```
max-radin, line 28: I think that `callable` is a function and not a type. Should be `Typing.Callable` I guess?

max-radin (changes requested):
See comment on `optimizer.py`.

src/python/zquantum/core/interfaces/optimizer.py
mstechly: Fixed in https://github.com/zapatacomputing/z-quantum-core/pull/211/commits/d2001de53c7fbe60d7bcf12d22f4dca1d3eb2990

max-radin (approved):

PR #212, opened 2021-02-23
177065997 move load specs

mstechly:


steps/circuit.py
```diff
@@ -16,19 +16,13 @@
     save_circuit_set,
 )
 from zquantum.core.testing import create_random_circuit as _create_random_circuit
-from zquantum.core.utils import create_object
+from zquantum.core.utils import create_object, load_from_specs
 import json
 from typing import Union, List, Optional, Dict
 
 Specs = Union[str, Dict]
```
simonwa7, line 23: @mstechly let's move the definition of the `Specs` type as well. Does that sound reasonable?

mstechly: Actually in another PR (https://github.com/zapatacomputing/z-quantum-core/pull/213) we're adding a `typing` module, I think this will be a more appropriate place to do so.

simonwa7: Okay, fair enough. Just going to reference this thread on that PR so that we do this before merging PR #213

simonwa7 (approved):

src/python/zquantum/core/utils.py
```diff
@@ -433,6 +433,7 @@ def create_object(specs, **kwargs):
     Returns:
         object: object of any type
     """
+    specs = copy.copy(specs)
```
simonwa7, line 436: Wait why are we removing this line? Without it, you can only "create" specs once which is a strange behavior.

PR #214, opened 2021-02-25
176997203 base hierarchy for unapplied gates

dexter2206:


mstechly (approved):
Looks good to me 👌

alexjuda (approved):
LGTM. I'd be nice to add docstrings to all classes and methods (like `Dagger` or `.controlled()`, but surely we can do it at the end when we're happy with the overall structure – especially after implementing the gate application representation.

PR #221, opened 2021-03-04
Graphs support

dlasecki:


steps/graph.py
```diff
@@ -33,14 +36,42 @@ def generate_random_regular_graph(
...
+        number_of_nodes: int, random_weights: bool = False, seed: Optional[int] = None
 ):
     graph = _generate_random_graph_erdos_renyi(
         number_of_nodes, 1.0, random_weights, seed
     )
     save_graph(graph, "graph.json")
 
 
+def generate_caveman_graph(
+        umber_of_cliques: int, size_of_cliques: int, random_weights: bool = False, seed: Optional[int] = None
```
mstechly, line 48:
```suggestion
        number_of_cliques: int, size_of_cliques: int, random_weights: bool = False, seed: Optional[int] = None
```

src/python/zquantum/core/graph_test.py
```diff
@@ -117,63 +120,131 @@ def test_generate_random_regular_graph(self):
...
+    def test_generate_caveman_graph(self):
+        # Given
+        number_of_cliques = 2
+        size_of_cliques = 3
+
+        # When
+        graph = generate_caveman_graph(number_of_cliques, size_of_cliques)
+
+        # Then
+        assert type(graph) == nx.Graph
```
mstechly, line 132: I would either change it to two separate tests – one for checking the type and the other for checking if the graph was created properly.

mstechly: Same for other tests.

steps/graph.py
```diff
@@ -9,10 +12,10 @@
 
 
 def generate_random_graph_erdos_renyi(
-    number_of_nodes: int,
-    edge_probability: float,
-    random_weights: bool = False,
-    seed: Optional[int] = None,
+        number_of_nodes: int,
```
mstechly, line 15: Looks to me like that's too much indentation.

mstechly (changes requested):

mstechly:
I've changed the target branch to `dev` and merged in latest changes.

src/python/zquantum/core/graph_test.py
dlasecki: I got rid of this. It does not seem to bring in much value anyway.

steps/graph.py
dlasecki: I think PEP wants an additional indentation when there is a line break.

mstechly: Fine 👍

mstechly (approved):

PR #222, opened 2021-03-08
Added conversion between Openfermion QubitOpenfermion and qiskti Weig…

mstechly:
…htedPauliOperator.

Tests are based on the equivalent pyquil tests.


src/python/zquantum/core/openfermion/_qiskit_conversions.py
```diff
@@ -0,0 +1,88 @@
...
+    Convert a OpenFermion QubitOperator to a WeightedPauliOperator.
+
+    Args:
+        qubit_operator: OpenFermion QubitOperator to convert to a qiskit.aqua.operators.WeightedPauliOperator
+
+    Returns:
+        WeightedPauliOperator representing the qubit operator
+    """
+    if not isinstance(qubit_operator, QubitOperator):
+        raise TypeError("qubit_operator must be a OpenFermion " "QubitOperator object")
```
max-radin, line 38: I think there are some legacy reasons why we were doing type checking. Should we remove this and rely on type hinting instead?

mstechly: I think one reason why we're doing it might be the fact that if you try to do it on `IsingOperator` it should work in principle but does not in practice and it's confusing as hell, so to make it more explicit we're throwing an error?
Or you think that was sth else?
Though it seems it should be working with `IsingOperator` as well 🤔

src/python/zquantum/core/openfermion/_qiskit_conversions.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if qubit_operator == QubitOperator():
```
max-radin, line 40: I think the conversion to qiskit won't behave as intended because `QubitOperator()` yields the zero operator, not the identity operator. The identity operator would be `QubitOperator(())`. Also, should we be using the same "I" representation for the identity term of an operator that also contains other terms?

max-radin (changes requested):
Before merging, we should fix the issue with `QubitOperator()` vs `QubitOperator(())`.

src/python/zquantum/core/openfermion/_qiskit_conversions.py
max-radin: I know there were cases historically where we had to distinguish `QubitOperator` from `IsingOperator`. But in this case it looks like the type checking came from from [forest-openfermion](https://github.com/rigetti/forest-openfermion/blob/53d0a9c7ebac1e540c793e3217593298d422aec2/forestopenfermion/pyquil_connector.py#L31-L33).

mstechly: Actually, another good reason to have this type check is because `count_qubits` doesn't work for `IsingOperator`. 
I did a PR, see here: https://github.com/quantumlib/OpenFermion/pull/712 .

src/python/zquantum/core/openfermion/_qiskit_conversions.py
mstechly: Fixed here: https://github.com/zapatacomputing/z-quantum-core/pull/222/commits/906d53e4f340993f4d9193f8c061281fbf093b85

src/python/zquantum/core/openfermion/_qiskit_conversions.py
```diff
@@ -0,0 +1,88 @@
...
+        QubitOperator representing the WeightedPauliOperator
+    """
+
+    if not isinstance(qiskit_pauli, WeightedPauliOperator):
+        raise TypeError("qiskit_pauli must be a qiskit WeightedPauliOperator")
+
+    transformed_term = QubitOperator()
+
+    for weight, qiskit_term in qiskit_pauli.paulis:
+        # openfermion_term_string_list = [f"[{term_pauli}{term_qubit}]" if term_pauli != "I" else ""
```
max-radin, line 73: Is there any reason to keep these commented out lines, or should they be deleted?

src/python/zquantum/core/openfermion/_qiskit_conversions.py
max-radin: If I'm not mistaken, the check for an empty operator isn't necessary since the rest of the code would still return the correct value.

max-radin (changes requested):
See comments.

src/python/zquantum/core/openfermion/_qiskit_conversions.py
mstechly: True! Fixed here: https://github.com/zapatacomputing/z-quantum-core/pull/222/commits/906d53e4f340993f4d9193f8c061281fbf093b85

mstechly: But still left the test, just in case ;)

src/python/zquantum/core/openfermion/_qiskit_conversions.py
mstechly: Good point!
Fixed here: https://github.com/zapatacomputing/z-quantum-core/pull/222/commits/8f0a6a5cd27d56286011c49bf04565b47b8cbeee

max-radin (approved):

PR #223, opened 2021-03-08
Update setup.py

kottmanj:
ruamel.yaml is necessary for inputs to succeed (imported over the pyquil dependencies)
problem might only occur when customImage is hosted


kottmanj:
It really only occurs on custom images. I better fix it there. Sorry

PR #227, opened 2021-03-09
Bitstring adds

martamau:
This PR is to merge two utils functions (+ tests) to load and save a List of BitstringDistribution objects.


src/python/zquantum/core/bitstring_distribution/_bitstring_distribution.py
```diff
@@ -164,6 +164,27 @@ def save_bitstring_distribution(
         f.write(json.dumps(dictionary, indent=2))
 
 
+
+def save_bitstring_distribution_set(
```
mstechly, line 168: `set` has [a pretty specific meaning](https://www.w3schools.com/python/python_sets.asp) in Python, so I guess `list` would be a much better name.
Also, when it comes to typing, it should be:

```
from typing import List
...
bitstring_distribution_set: List[BitstringDistribution]
```

mstechly (changes requested):

src/python/zquantum/core/bitstring_distribution/_bitstring_distribution.py
mstechly: Same elsewhere :)

martamau: Agree for the typing (I actually just added the tests as the rest was implemented already), fixed [here](https://github.com/zapatacomputing/z-quantum-core/pull/227/commits/c850d4e27c8ea45739a89278b29fac6c67d7e091). 
As for the name, I see your point but in all the other cases we have save_circuit_set, save_operator_set etc etc, so it might make sense to be consistent? Not sure about it

mstechly: That's a good point 🤔 
Let's go ahead then.

mstechly (approved):

PR #228, opened 2021-03-12
Fixing bug and changing measurement generation from frequencies

jgonthier:
This is fixing a bug in the generation of measurements from bitstring frequencies: in the leftover samples, each bitstring was added only once, even if it was sampled multiple times.

However, I also changed how the measurements were generated to obtain something as close as possible to the original distribution: now the measurements are generated by rounding `frequency * number_of_samples` instead of simply using `int()` which is equivalent to `int(floor())`. Then, that means that if frequencies are inconsistent with the number of samples desired, we can sometimes get more and sometimes less samples than desired.

The bitstrings to correct are chosen by sampling, using the non-integer part of `frequency * number_of_samples` for each, but giving more weight when the non-integer part is closest to 0.5. The reasoning is that 2.9 measurements should be rounded to 3, 2.1 to 2, and if a rounding error occurs it's likely to be for something like 2.5002 or 2.499.

With that, all the original tests pass, and I also added a new test that would fail with the bug corrected in this PR.


max-radin:
Ah thanks for finding and fixing this!

One reason we had originally gone with the floor method was that @pediejo pointed out that it is unbiased in the sense that that if you called `get_measurements_representing_distribution` over and over again, the frequencies of each bitstring would approach the desired distribution.

On the other hand the rounding method is biased, for example if your residuals are `[0.6, 0.4]` then the first bitstring will always be overrepresented and the second underrepresented. It seems reasonable though to prefer a biased set of measurements that's closer to the target distribution (or at least give users this option).

Another biased approach we had considered was just to distribute the leftovers to the bitstrings with the highest residuals, one shot for each bitstring. @jgonthier @pediejo do you guys have a sense of how this roundrobin method compares to the rounding method, or how much we care about the bias? (Also, I wouldn't want to spend too much energy on optimizing this because this is all just a workaround to deal with the fact that the estimator interface doesn't support bitstring distributions.)

jgonthier:
Thanks for the background on the floor method! I think there are actually two different possible uses here:
1. We want to sample various number of bitstrings from an existing distribution 
2. We want to reconstruct the original measurements from an experimentally obtained distribution

In case 2., there should actually not be residuals because each frequency in the distribution is N_{bitstring} / N_{tot}, we know N_{tot} and we want to reconstruct N_{bitstring}.
In case 1., the user can ask for any number of measurements and residuals are a potential issue.

All of that to say that I think it would make sense to give users the option between 1. and 2.

I feel like the roundrobin method may combine the worst aspects of both the rounding method and the floor method in that it may be biased and further from the original distribution in the case of an experiment, but I actually don't have any proof regarding the bias, it's just a feeling.

max-radin (approved):

PR #233, opened 2021-03-17
Circuits convenience api improvements

dexter2206:
This reimplements how custom gate definitions are handled. The bottom line: gate definitions don't have to be passed separately as argument to Circuits initializer.


src/python/zquantum/core/wip/circuits/_circuit.py
```diff
@@ -1,9 +1,11 @@
 from functools import singledispatch, reduce
+import operator
 from typing import Union, Dict, Optional, Iterable, Any
 
 import sympy
 
-from . import _gates
+from . import _gates, MatrixFactoryGate
```
alexjuda, line 7: You're importing `MatrixFactoryGate` from the `__init__` (I suppose PyCharm generated this import), did you consider importing the symbol from the module it was originally defined in? 

My gut feeling is that the latter approach would be more coherent with the scenarios where we import symbols not exposed in the public API (probably the majority of them), but I'm happy to hear your thoughts.

alexjuda (changes requested):
Nice, good idea with that lazy custom gate introspection deferred to the time when we need it. I had a different solution in mind, but I've figured out yours is better.

src/python/zquantum/core/wip/circuits/_circuit.py
dexter2206: I agree, and indeed, the import was autogenerated. I addressed this in 9e1bdf9.

alexjuda (approved):

PR #234, opened 2021-03-19
Fix error when printing circuit in a repl

alexjuda:
I stumbled opon it when I prepared the usage doc. It's a leftover from previous implementation when `Circuit` had `custom_gate_definitions` field, now it's computed dynamically (introduced in #233).

### DIFF
--- a/src/python/zquantum/core/wip/circuits/_circuit_test.py
+++ b/src/python/zquantum/core/wip/circuits/_circuit_test.py
@@ -62,7 +62,7 @@ def test_creating_circuit_has_correct_operations():
     Circuit([H(0)]),
     Circuit([H(0)], 5),
 ])
-def test_printing_circuit_doesnt_raise(circuit):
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/wip/circuits/_circuit_test.py
```diff
@@ -55,6 +56,17 @@ def test_creating_circuit_has_correct_operations():
     assert circuit.operations == list(EXAMPLE_OPERATIONS)
 
 
+@pytest.mark.parametrize("circuit", [
+    Circuit(),
+    Circuit([]),
+    Circuit([H(0)]),
+    Circuit([H(0)], 5),
+])
+def test_printing_circuit_doesnt_raise(circuit):
```
mstechly, line 65: I would say name it `doesnt_raise_exception` , but I think it's a matter of taste ;)

mstechly (approved):

PR #235, opened 2021-03-19
Created docker image 0.0.7.

mstechly:


mstechly:
Updated some libraries and added `codecarbon` and `cvxpy`.
The latter is used in `z-quantum-qubo` now and I added it because it takes a lot of time to install it which increases startup time.

docker/z-quantum-default/0.0.7/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+                cd qvm && \
+                git fetch && \
+                git checkout v1.17.1 && \
+                make QVM_WORKSPACE=10240 qvm && \
+                mv qvm /usr/local/bin
+
+RUN python3 -m pip install rpcq==3.6.0 \
+                    pytest==5.3.5 \
+                    networkx==2.4 \
+                    pyquil==2.25.0 \
```
simonwa7, line 59: Should we actually downgraded pyquil from `2.27.0` to `2.25.0`? I think it would be better to keep it upgraded no?

simonwa7:
Additionally, we use a lot of `==` for pip versioning. Might it be better to use the same "loose" versioning as in `setup.py`?

docker/z-quantum-default/0.0.7/Dockerfile
mstechly: For some reason we have the following in the setup:
https://github.com/zapatacomputing/z-quantum-core/blob/45aefee4eea05bce8d26c201bd471b76d05db139/setup.py#L28

mstechly: So I've changed the docker image accordingly.

mstechly:
I think that actually for the docker image we want the version to be very specific.
If they need to be overwritten later, that's fine.

simonwa7:
Oh fair point. Perhaps this isn't an issue, but I'm just wondering out loud here:
We keep loose bounds in the `setup.py` so that packages will get regular updates. Subsequently, we will also get those updates and then be able to identify when an update causes something to break. However, if we keep strict bounds in our Dockerfile, we'll never "see" those updates reflected at runtime since the `setup.py` will accept the environment of the image. 

This means that if we want to keep strict bounds (`==`) we would either need to: 
- periodically update the `setup.py` to force newer versions (which is contrary to trying to decrease the constraints)
- or periodically update the Dockerfile to use newer versions

mstechly:
I think we're rarely updating docker images, so if we just stick to updating them to whatever is specified by the `setup.py` at the time we want to make an update, it should work.
This way we always update to the version that we already know it's working.
Is that what you had in mind?

simonwa7:
I agree that it will always work in the sense that we'll always have a docker image with an environment compatible with z-quantum-core. My question/comment is not so much regarding "how do we ensure the docker image will work" and more regarding "how do we ensure the bounds in our `setup.py` are accurate and that we continue to get minor fixes/improvements in the underlying packages". 

For example, say we set the version of package "`quantum-simulator`" to `>=0.2.0` because we know everything works with that version. In our dockerfile, if we set `quantum-simulator==0.2.0`, then we will _always_ get that version. And never get `0.2.1` or other "improved" versions. I see two issues with this:
1. Say a new version comes out. It contains a minor quality of life improvement: the "quantum simulator" in the package gets a 1.2x speedup. We want this minor improvement, however, we want to keep the bounds in `setup.py` "loose" enough so that if another package conflicted with `0.2.1`, then it could still work.
2. Say a new version comes out, but this time it includes a change that breaks something for us. Since we have a strict bound in the Dockerfile, we'll never see this breaking change on Orquestra and therefore we wouldn't get alerted of the change. Yet in `setup.py` the bounds would indicate that `0.2.1` is an "acceptable" version (since we say `>=0.2.0`).

mstechly:
I see... Great points, but no idea how to deal with that.
Through a CICD perhaps?
Anyway – can we move this discussion elsewhere and merge this ;) ?

simonwa7 (approved):

alexjuda:
I agree with @mstechly's point, building docker image is *the* situation where we usually want to have locked dependencies, simply to provide reproducible builds. 

I also agree with @simonwa7 that manual maintenance of dependency pins is a hassle. But this doesn't mean it can't be improved, this is exactly why there are tools like [pip-tools](https://github.com/jazzband/pip-tools), or the newer dependency managers like [pipenv](https://github.com/jazzband/pip-tools) or [poetry](https://github.com/python-poetry/poetry) that almost abstract that away completely. Perhaps going this direction in the future (using some form of a lockfile) could alleviate some pain around the `==` constraint management?

PR #236, opened 2021-03-23
177434713 Gates parameter replacement

dexter2206:
This implements replacing parameters in gates and gate operations. In particular one can:

- replace parameters in instances of `MatrixFactoryGate`
- replace parameters in instances of modified gates: `ControlledGate` and `Dagger`
- replace parameters in instances of `GateOperation`.

Additionally, binding parameters is implemented on `GateOperation` level, so that one does not have to construct its instances explicitly.

Note that "replacing" here means non-mutating replacement, i.e. producing new instances with given parameters override.


src/python/zquantum/core/wip/circuits/_gates_test.py
```diff
@@ -175,3 +193,39 @@ def test_binding_parameters_in_control_gate_is_propagated_to_wrapped_gate(
...
+
+    def test_binding_parameters_constructs_operation_of_gate_with_bound_parameters(
+        self, gate
+    ):
+        op = GateOperation(gate, tuple(range(gate.num_qubits)))
+        symbols_map = {sympy.Symbol("phi"): 0.5, sympy.Symbol("y"): 1.1}
+        assert (
+            op.bind(symbols_map) ==
+            GateOperation(gate.bind(symbols_map), tuple(range(gate.num_qubits)))
+        )
```
alexjuda, line 220: This assertion duplicates implementation of the code it's supposed to verify. WDYT about changing this test's assertion to a one that would reflect the test case name?
```suggestion
        assert set(op.bind(symbols_map).gate.params) == set(symbols_map.values())
```

src/python/zquantum/core/wip/circuits/_gates_test.py
```diff
@@ -175,3 +193,39 @@ def test_binding_parameters_in_control_gate_is_propagated_to_wrapped_gate(
...
+    def test_replacing_parameters_constructs_operation_of_gate_with_new_parameters(
+        self, gate
+    ):
+        op = GateOperation(gate, tuple(range(gate.num_qubits)))
+        new_params = tuple(-1 * param for param in op.params)
+
+        assert (
+            op.replace_params(new_params) ==
+            GateOperation(gate.replace_params(new_params), tuple(range(gate.num_qubits)))
+        )
```
alexjuda, line 231: Same question as above

alexjuda (changes requested):

alexjuda (approved):

PR #237, opened 2021-03-23
Adding S gate to built in gates

simonwa7:


simonwa7:
As a user-experience note, it seems a _bit_ confusing how to add the gate definition to the conversion packages. It took me awhile to figure out how it was working in `pyquil` since there's no explicit mapping.

I also noticed that after I added the definition, but before I modified the conversions nothing failed. I'm wondering if that's because the conversions were "smart" enough to handle it (or perhaps create a custom gate) or if it just wasn't being tested? It would be good to have clarification on this @dexter2206 @alexjuda

alexjuda (approved):
👍

alexjuda:
> As a user-experience note, it seems a _bit_ confusing how to add the gate definition to the conversion packages. It took me awhile to figure out how it was working in `pyquil` since there's no explicit mapping.

Do you have any ideas besides including a chapter this topic in the documentation? At the moment we don't have separate docs, so my idea was to include it as a module docstring. Do you think that would be helpful?

> I also noticed that after I added the definition, but before I modified the conversions nothing failed. I'm wondering if that's because the conversions were "smart" enough to handle it (or perhaps create a custom gate) or if it just wasn't being tested? It would be good to have clarification on this @dexter2206 @alexjuda

This probably should be stated more explicitly somewhere, but I think this is exactly what happened – `pyquil_conversions` module is smart enough to assume that ZQuantum and PyQuil have very similar set of built-in gates, so it defaults to get the correspoding gate just by the input one's name.

simonwa7:
> Do you have any ideas besides including a chapter this topic in the documentation? At the moment we don't have separate docs, so my idea was to include it as a module docstring. Do you think that would be helpful?

I think a module docstring would be appropriate that way for each conversion library, it's clear what to do to make it accessible.

> This probably should be stated more explicitly somewhere, but I think this is exactly what happened – pyquil_conversions module is smart enough to assume that ZQuantum and PyQuil have very similar set of built-in gates, so it defaults to get the correspoding gate just by the input one's name.

That was my hunch, but it was also the case for `cirq` and `qiskit` too as I hadn't modified them either. I think it would also be good to clarify (for each module, probably in the module docstring again) whether the converter is smart enough to assign the gate to the proper native-gate without explicitly declaring it (ie. zquantum S gate would be converted to qiskit S gate) or whether it's building a custom gate if it isn't defined (ie. zquantum S gate gets converted to custom qiskit gate).

alexjuda:
Added as a comment under the story for new circuits API docs.

PR #238, opened 2021-03-24
Prepare for PyPI

alexjuda:
Updates setup.py for releasing artifacts on PyPI. [This](https://pypi.org/project/zquantum-core/) is how the releases look like.

Note the package name change. With my changes the users would type `pip install zquantum-core`, which is nice, because it corresponds to the import name (`import zquantum.core`). 

If we don't introduce my change, users will have to run `pip install z-quantum-core`, but the import name would be different (`import zquantum.core`), and I'd like to avoid confusion like that if we still can.


alexjuda:
After a discussion on Slack I've rolled back the package rename. [This](https://pypi.org/project/z-quantum-core/) is the PyPI release page, [here](https://test.pypi.org/project/z-quantum-core/)'s the test one.

dexter2206 (approved):

PR #241, opened 2021-03-26
Create main.yml

mstechly:


simonwa7 (approved):

mbeach42:
@mstechly 

Black formatter fails due to line 131 in https://github.com/zapatacomputing/z-quantum-core/blob/master/src/python/zquantum/core/wip/circuits/conversions/cirq_conversions_test.py

Just change the __unitary__ to 
```python
def _unitary_(self):
        return (
            np.array(
                [
                    [np.cos(self.theta), np.sin(self.theta)],
                    [np.sin(self.theta), -np.cos(self.theta)],
                ]
            )
            / np.sqrt(2)
        )
        ```

File: .github/workflows/style.yml [hunk 1]
```diff
@@ -0,0 +1,62 @@
+name: style
+
+# Controls when the action will run. 
+on:
+  push:
+    branches: [ master, dev ]
```
mbeach42, line 6: let's add `main` here too. Some repos might use main instead of master in the future.

.github/workflows/style.yml
```diff
@@ -0,0 +1,62 @@
...
+          # A list of all paths to test
+          python-root-list: src/
+          # Use Pylint
+          use-pylint: false
+          # Use pycodestyle
+          use-pycodestyle: false
+          # Use Flake8
+          use-flake8: true
+          # Use Black
+          use-black: true
```
mbeach42, line 46: does black run before flake8 or after? I would assume after is better.

mstechly: Black is check only anyway so I don't think it matters

.github/workflows/style.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           # Extra options: pycodestyle $(extra-pycodestyle-options) $(python-root-list)
           # extra-pycodestyle-options: # optional, default is 
           # Extra options: flake8 $(extra-flake8-options) $(python-root-list)
-          extra-flake8-options: --ignore=E203,E266,F401
+          extra-flake8-options: --ignore=E203,E266,F401 --max-line-length=88
```
mbeach42, line 56: 👍 :) thanks! fewer errors now!

PR #242, opened 2021-03-30
Docs for reworked circuits

alexjuda:
We don't have sphinx docs set up yet so I've added new circuits usage guide to the package docstring.


src/python/zquantum/core/wip/circuits/__init__.py
```diff
@@ -1,4 +1,116 @@
...
+
+
+Defining new gates
+------------------
+
+To use a gate that isn't already covered by built-in ones you can define a custom gate
+or extend the set of the built-in ones and file a PR to z-quantum-core.
+
+Using custom gates::
+    CustomA = circuits.CustomGateDefinition(
```
mstechly, line 72: I would change it to `custom_A` as it's a more pythonic name of the variable.

mstechly (approved):

src/python/zquantum/core/wip/circuits/__init__.py
alexjuda: Good point! Updated in 21c28f3.

PR #244, opened 2021-03-31
Refactored optimal shot allocation as a separate function

jgonthier:
This allows other estimators to easily use optimal shot allocation without duplicating code.

### DIFF
--- a/src/python/zquantum/core/estimator.py
+++ b/src/python/zquantum/core/estimator.py
@@ -110,13 +110,6 @@ def allocate_shots(
     Returns:
         List of integers giving the number of shots for each frame operator.
     """
-    if shot_allocation_strategy not in (
-        "optimal",
-        "uniform",
-    ):
-        raise ValueError(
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/estimator.py
```diff
@@ -84,6 +84,67 @@ def get_context_selection_circuit_for_group(
...
+        prior_expectation_values: If given, this is used to estimate the variances of the frame operators when
+                                  doing optimal shot allocation.
+    Returns:
+        List of integers giving the number of shots for each frame operator.
+    """
+    if shot_allocation_strategy not in (
+        "optimal",
+        "uniform",
+    ):
+        raise ValueError(
```
mstechly, line 117: Another way of doing it would be to put this raise in the`else` statement after `if shot_allocation_strategy == "uniform":` and `if shot_allocation_strategy == "optimal":` .
Though I'm not sure if that's a definitely better solution, just different, so it can stay the way it is :)

mstechly (approved):

PR #247, opened 2021-03-31
Setup file test dependancies

mbeach42:
separate test dependancies in setup.py.
Also updated README to reflect how to install with the test dependancies for development


README.md
```diff
@@ -5,23 +5,24 @@
...
 In order to use `z-quantum-core` in your workflow, you need to add it as an `import` in your Orquestra workflow:
 
 ```yaml
 imports:
-- name: z-quantum-core
-  type: git
-  parameters:
-    repository: "git@github.com:zapatacomputing/z-quantum-core.git"
-    branch: "master"
+  - name: z-quantum-core
```
alexjuda, line 21: was this indentation intended? TBH I like this way of writing `.yaml`s better, but I recall in our docs we usually go with `-` at the container level

setup.py
```diff
@@ -2,10 +2,23 @@
...
+    "pytest>=3.7.1",
+    "pytest-cov>=2.5.1",
+    "tox>=3.2.1",
+    "flake8>=3.7.9",
+    "black>=19.3b0",
+    "pre_commit>=2.10.1",
+]
+
+extras_require = {
+    "test": test_requires,
```
alexjuda, line 15: WDYT about using a `dev` extra instead of `test`? This would be more in line with other package managers like [poetry](https://python-poetry.org/docs/cli/#options_3) or [npm](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file)

alexjuda (changes requested):
Thanks for finally separating the development deps! I couldn't wait for this moment to happen  😄

setup.py
mbeach42: yes, I would like to separate `test` and `dev`. Something like this:
```python
dev_requires = test_requires + ["pre_commit>=2.10.1", "black"]

extras_require = {
    "dev": dev_requires,
    "test": test_requires,
    "docs": coverage_requires,
}
```

How do people feel about this? @zapatacomputing/quantum-software

README.md
mbeach42: Just automatic formatting at work here. If we establish prettier as the yaml formatter, this would be the new style.

setup.py
mstechly: Not sure why `"docs": coverage_requires` and not `"docs": docs_requires`, but in general this looks fine to me.

mbeach42: @mstechly oops! I had separate docs, build, test, coverage, dev originally. It just just an example from an old project :P

alexjuda: I agree it's a good idea to separate dependencies needed when we develop the library from the ones that a user needs to install when they only want to use the library in their project. 

Going farther with the granularity than just two options seems counter-productive for me, because this would require me to specify all the extras when I develop the library. Is there a reason we don't wanna do the same path as poetry and npm?

alexjuda: tl;dr my suggestion is to use just a single, `dev` extra that would contain all development tools, like `pytest` or `black`

mbeach42:
moved to the pre-commit PR

PR #249, opened 2021-03-31
Removed unused import, and ran sort 

mbeach42:
This is the result of running the first two parts of a `.pre-commit-config.yml` file 
```
repos:
  # automatically removes unused imports
  - repo: https://github.com/myint/autoflake.git
    rev: v1.4
    hooks:
      - id: autoflake
        args:
          [
            --in-place,
            --expand-star-imports,
            --ignore-init-module-imports,
            --remove-all-unused-imports,
            --remove-unused-variables,
            --remove-duplicate-keys,
          ]

  - repo: https://github.com/PyCQA/isort
    rev: 5.8.0
    hooks:
      - id: isort
        args: ["--profile", "black", "--filter-files"]
```
then `pre-commit run --all-files`


src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -1294,7 +1288,7 @@ def test_to_quil(self):
         gate_H0 = Gate("H", [qubits[0]])
         gate_CNOT01 = Gate("CNOT", [qubits[0], qubits[1]])
         gate_T2 = Gate("T", [qubits[2]])
-        gate_CZ12 = Gate("CZ", [qubits[1], qubits[2]])
+        Gate("CZ", [qubits[1], qubits[2]])
```
mstechly, line 1291:
```suggestion
```

src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -1307,7 +1301,7 @@ def test_to_text_diagram(self):
...
```
mstechly, line 1304:
```suggestion
```

src/python/zquantum/core/evolution_test.py
```diff
@@ -109,7 +109,7 @@ def test_time_evolution_derivatives(self):
         )
         time_symbol = sympy.Symbol("t")
         time_value = 0.4
-        symbols_map = [(time_symbol, time_value)]
+        [(time_symbol, time_value)]
```
mstechly, line 112: Looks like the logic in this test needs some fixing, as `time_value` is not actually needed.

src/python/zquantum/core/history/recorder.py
```diff
@@ -143,31 +145,27 @@ def recorder(
     save_condition: SaveCondition = always,
 ) -> ArtifactRecorderWithGradient:
     """The recorder function: variant for callables with gradient and storing artifacts."""
-    pass
```
mstechly, line 146: I'm not 100% whether it will work without `pass` .
Also I think that having them here is more explicit and increases readability.

src/python/zquantum/core/interfaces/ansatz_test.py
```diff
@@ -15,7 +14,7 @@ def test_set_number_of_layers_invalidates_parametrized_circuit(self, ansatz):
         # Given
         new_number_of_layers = 100
         if ansatz.supports_parametrized_circuits:
-            initial_circuit = ansatz.parametrized_circuit
+            ansatz.parametrized_circuit
```
mstechly, line 17: I prefer this syntax, as it's more readable to me
```suggestion
            _ = ansatz.parametrized_circuit
```

src/python/zquantum/core/circuit/_circuit_test.py
```diff
@@ -253,7 +248,6 @@ def test_circuit_evaluate_with_wrong_params(self):
         # Given
         theta_1 = sympy.Symbol("theta_1")
         theta_2 = sympy.Symbol("theta_2")
-        value_1 = 0.5
         value_2 = 0.6
         symbols_map = [(theta_2, value_2)]
```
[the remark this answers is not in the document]
alexjuda: @mbeach42 and I agreed that we should keep the tests unchanged for now, we can revisit this refactor later

src/python/zquantum/core/evolution_test.py
alexjuda: Let's do the refactor in another PR

src/python/zquantum/core/history/recorder.py
alexjuda: Python parser needs `pass` only in situations where a block has no other body. Here we have a docstring, so the parser sees the indentation, so we don't need the `pass`

mstechly: Good to know :)

src/python/zquantum/core/wip/circuits/_serde.py
```diff
@@ -1,13 +1,10 @@
...
+from typing import Iterable, List
 
 import sympy
 
-from . import _gates
-from . import _builtin_gates
-from . import _circuit
 from ...utils import SCHEMA_VERSION
-
+from . import _builtin_gates, _circuit, _gates
```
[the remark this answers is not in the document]
alexjuda: If I was to write this by hand I would probably keep the imports in separate lines, but I'm fine with the autoformatter merging them. This doesn't change the logic so it's just esthetics.

src/python/zquantum/core/circuit/_circuit_test.py
alexjuda: Fixed in 71301c5

src/python/zquantum/core/circuit/_circuit_test.py
alexjuda: Updated in 71301c5

src/python/zquantum/core/interfaces/ansatz_test.py
alexjuda: I've removed the line in 71301c5, assuming that we shouldn't need to test property getters.

mstechly (approved):

PR #251, opened 2021-04-01
Run black on the whole repo

alexjuda:
I hope this will be easier than manually fixing linter errors that cause PR check failures in #250.


alexjuda:
Seems like it doesn't fix all the linter issues I intended to. I'll close it for now not to mess with other hanging PRs that touch a lot of codebase.

PR #256, opened 2021-04-07
Typing fixes

max-radin:
This fixes mypy errors in a few different files.


src/python/zquantum/core/bitstring_distribution/distance_measures/mmd.py
```diff
@@ -62,14 +62,13 @@ def compute_multi_rbf_kernel(x_i: np.array, y_j: np.array, sigmas: List) -> np.n
...
     exponent = np.abs(x_i[:, None] - y_j[None, :]) ** 2
-    kernel_matrix = 0.0
     for sigma in sigmas:
         try:
             gamma = 1.0 / (2 * sigma)
         except ZeroDivisionError as error:
             print("Handling run-time error:", error)
             raise
-        kernel_matrix = kernel_matrix + np.exp(-gamma * exponent)
+        kernel_matrix = np.exp(-gamma * exponent)
```
mstechly, line 71: I think you have changed the logic here. Shouldn't it be `+=`?

src/python/zquantum/core/measurement.py
```diff
@@ -659,7 +674,11 @@ def concatenate_expectation_values(
         The combined expectation values.
     """
 
-    combined_expectation_values = ExpectationValues(np.zeros(0,))
+    combined_expectation_values = ExpectationValues(
```
mstechly, line 677: Actually, it'll be cleaner if you just remove `,` after 0 and blackify it :)

mstechly (changes requested):

src/python/zquantum/core/bitstring_distribution/distance_measures/mmd.py
max-radin: Ah you're right. I had seen that `kernel_matrix` was being initialized to zero, but didn't see that the addition was happening in a loop.

alexjuda: fixed in 2d5abfa

src/python/zquantum/core/measurement.py
alexjuda: Done in d8682df...2c47b89

src/python/zquantum/core/bitstring_distribution/distance_measures/mmd.py
mstechly: Power of PRs!

mstechly (approved):

mstechly:
Ready to go after conflicts are resolved.

PR #259, opened 2021-04-08
Draft of new Estimator interface.

mstechly:
The structure of the functions inside the module might still change in future.


mstechly:
@max-radin I wasn't sure about the commented part in `naively_estimate_expectation_values` – could you please check whether it should or should not be uncommented?

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,236 @@
...
+from ...interfaces.backend import QuantumBackend, QuantumSimulator
+from ...measurement import (
+    ExpectationValues,
+    expectation_values_to_real,
+    concatenate_expectation_values,
+)
+
+greedy_grouping_with_context_selection: EstimationTaskTransformer
+uniform_shot_allocation: EstimationTaskTransformer
+proportional_shot_allocation: EstimationTaskTransformer
```
alexjuda, line 23: what's the purpose of these lines?

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def get_context_selection_circuit(
+    operator: QubitOperator,
+) -> Tuple[Circuit, IsingOperator]:
+    """Get the context selection circuit for measuring the expectation value
+    of a Pauli term.
+    TODO
```
alexjuda, line 31: looks like an overlooked TODO :)

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,236 @@
+from typing import List, Tuple, Optional
+from openfermion import QubitOperator, IsingOperator
+import pyquil
+import numpy as np
+
+from ...circuit._circuit import Circuit
```
alexjuda, line 6: Did you consider using the reworked circuits here?

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,236 @@
...
+
+def uniform_shot_allocation(number_of_shots: int) -> EstimationTaskTransformer:
+    """
+    Returns an EstimationTaskTransformer which allocates the same number of shots to each task.
+
+    Args:
+        number_of_shots: number of shots to be assigned to each EstimationTask
+
+    Returns:
+        EstimationTaskTransformer
```
alexjuda, line 128: We're using [Google's docstring convention](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy), right? If so, then the type should be only specified in the type annotation and not in the docstring.

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,236 @@
...
+from ...circuit._circuit import Circuit
+from .estimation_interface import (
+    EstimationTask,
+    EstimationTaskTransformer,
+)
+from ...hamiltonian import group_comeasureable_terms_greedy, estimate_nmeas_for_frames
+from ...utils import scale_and_discretize
+from .estimation_interface import EstimationTask
+from ...interfaces.backend import QuantumBackend, QuantumSimulator
+from ...measurement import (
```
alexjuda, line 15: Did you consider using absolute imports? This way we wouldn't have to change them after we move these files away from the `wip` package.

src/python/zquantum/core/wip/estimators/estimation_interface.py
```diff
@@ -0,0 +1,44 @@
+from ...circuit import Circuit
+from ...interfaces.backend import QuantumBackend
+from ...measurement import ExpectationValues
+from openfermion import SymbolicOperator
+from typing import List, Callable
+from dataclasses import dataclass
+from typing_extensions import Protocol
+
+
+@dataclass
```
alexjuda, line 10: Did you consider using an immutable dataclass?
```suggestion
@dataclass(frozen=True)
```

tests/zquantum/core/wip/estimators/estimation_test.py
```diff
@@ -0,0 +1,198 @@
...
+        frame_operators,
+        total_n_shots,
+        prior_expectation_values,
+        target_n_samples_list,
+    ):
+        allocate_shots = proportional_shot_allocation(
+            total_n_shots, prior_expectation_values
+        )
+        estimation_tasks = [
+            EstimationTask(operator, None, None) for operator in frame_operators
```
alexjuda, line 117: I believe this would cause mypy complaints - passing `None` to non-optional fields. Do I understand correctly is that the `None`s here are because the fields don't matter for the use case in this test? If such use cases exist shouldn't we change the `EstimationTask`, e.g. by making the 2nd & 3rd field optional?

alexjuda (changes requested):
I've left some questions regarding the code style. I'd be nice if another person also validated the domain-specific logic cc @dexter2206 @simonwa7 @max-radin

src/python/zquantum/core/wip/estimators/estimation.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+    measurements_list = backend.run_circuitset_and_measure(circuits, shots_per_circuit)
+
+    expectation_values_list = [
+        expectation_values_to_real(measurements.get_expectation_values(frame_operator))
+        for frame_operator, measurements in zip(operators, measurements_list)
+    ]
+
+    # TODO handle empty term?
```
max-radin, line 203: The reason this logic was included in the `BasicEstimator` was that the grouping functions [throw away the constant term](https://github.com/zapatacomputing/z-quantum-core/blob/943f28bb23f404ad1f1c7c2f68ce3656993e692d/src/python/zquantum/core/hamiltonian.py#L58-L59). I think that, as currently written, `greedy_grouping_with_context_selection` + `naively_estimate_expectation_values` will yield the wrong answer because the constant term will get lost.

One way to address this could be to just implement [the story about making the constant term included everywhere](https://www.pivotaltracker.com/story/show/176774633). I think the needed changes are:
* Change `group_comeasureable_terms_greedy` to keep the constant term
* Update [the check for the constant term in `compute_group_variances`](https://github.com/zapatacomputing/z-quantum-core/blob/943f28bb23f404ad1f1c7c2f68ce3656993e692d/src/python/zquantum/core/hamiltonian.py#L122-L125
)
* Remove [the logic to exclude the constant term](https://github.com/zapatacomputing/z-quantum-core/blob/d2da45b051d0460cffcd6351470e23c0458fa3dd/src/python/zquantum/core/hamiltonian.py#L206-L209) from `get_expectation_values_from_rdms`.

What do you think about trying to implement the above changes in a separate PR?

src/python/zquantum/core/wip/estimators/estimation.py
mstechly: Fixed in https://github.com/zapatacomputing/z-quantum-core/pull/259/commits/c41fe3f9eebe12e059977a0d6cffecf9ee42865b

src/python/zquantum/core/wip/estimators/estimation.py
mstechly: Sounds good, will make a PR to this branch.

src/python/zquantum/core/wip/estimators/estimation.py
alexjuda: FYI: the original point of Python's protocols was that you don't need to explicitly mark the code that conforms to a one, you can just use the protocol as a type hint on the use site and the type checker would still utilize that. However, explicit marks can still make sense for the sake of documentation and indeed this makes sense here IMHO.

tests/zquantum/core/wip/estimators/estimation_test.py
alexjuda: Thanks, testing using data that's close to the data that would be used in production is usually a good thing IMO - just like in ML, this guards us against unknown bias 🙂

alexjuda (approved):

src/python/zquantum/core/wip/estimators/estimation.py
dexter2206: @mstechly We determined that it shouldn't be there, as it trips type checkers. Unfortunately, there seems to be no way for marking functions implementing protocols explicitly.

mstechly: That's what Code Review is for :D 
Fixed in https://github.com/zapatacomputing/z-quantum-core/pull/259/commits/a65bd21d57a88fcd08eebfe7e826aa99b86e6792

PR #261, opened 2021-04-09
Adding simple fix for bug when checking is bitstring distribution on …

simonwa7:
…object that will cause IndexError


src/python/zquantum/core/bitstring_distribution/_bitstring_distribution.py
```diff
@@ -101,11 +101,14 @@ def is_bitstring_distribution(input_dict: Dict) -> bool:
...
-        and is_key_length_fixed(input_dict)
-        and are_keys_binary_strings(input_dict)
-    )
+    try:
+        return (
+            is_non_negative(input_dict)
+            and is_key_length_fixed(input_dict)
+            and are_keys_binary_strings(input_dict)
+        )
+    except IndexError:
```
alexjuda, line 110: Is there any other possibility for raising `IndexError` other than by passing `input_dict={}`? If that's the only case, WDYT about handling it explicitly, e.g. with another predicate? Like
```
return (
         is_non_empty(input_dict)
         and is_non_negative(input_dict)
         and is_key_length_fixed(input_dict)
         and are_keys_binary_strings(input_dict)
     )
```

alexjuda, line 110: Is there any other possibility for raising `IndexError` other than by passing `input_dict={}`? If that's the only case, WDYT about handling it explicitly, e.g. with another predicate? Like
```
return (
         is_non_empty(input_dict)
         and is_non_negative(input_dict)
         and is_key_length_fixed(input_dict)
         and are_keys_binary_strings(input_dict)
     )
```

alexjuda (changes requested):

src/python/zquantum/core/bitstring_distribution/_bitstring_distribution.py
simonwa7: Agreed, this solution is better. Integrated in 948f1b9

alexjuda (approved):

PR #265, opened 2021-04-14
Translate old circuits to new ones

alexjuda:
Required by @dexter2206's compatibility layer decorator.

Commits done during pairing with @dexter2206.


src/python/zquantum/core/wip/circuits/_compat.py
```diff
@@ -0,0 +1,7 @@
+import zquantum.core.circuit as old_circuit
```
mstechly, line 1: Can we name the file `compatiblity` and not `compat` :D ?

mstechly (approved):

src/python/zquantum/core/wip/circuits/_compat.py
alexjuda: Sure 🙂 (see 833bf7c)

PR #266, opened 2021-04-14
Quantum compiling

simonwa7:


mstechly:
Haven't looked too hard, but I'm wondering with `z-quantum-vqe` wouldn't be a better place?

max-radin (approved):

PR #269, opened 2021-04-15
Update create object

simonwa7:


mstechly (approved):
Looks good to me, just make sure it's tested with a couple of actual workflows that use these functions.
Cause if something doesn't work it will affect everyone ;)

src/python/zquantum/core/utils.py
```diff
@@ -440,8 +443,26 @@ def create_object(specs: Dict, **kwargs):
...
-    created_object = creator(**specs, **kwargs)
-    return created_object
+
+    for key in specs.keys():
+        if key in kwargs.keys():
+            raise ValueError("Cannot have same parameter assigned to multiple values")
+
+    if isinstance(creator, FunctionType):
+        if kwargs != {} or specs != {}:
+            import inspect
```
alexjuda, line 453: Is there a reason why the import is here and not on the top of the file?

src/python/zquantum/core/utils.py
```diff
@@ -440,8 +443,26 @@ def create_object(specs: Dict, **kwargs):
...
     creator_name = specs.pop("function_name")
     creator = getattr(module, creator_name)
-    created_object = creator(**specs, **kwargs)
-    return created_object
+
+    for key in specs.keys():
+        if key in kwargs.keys():
+            raise ValueError("Cannot have same parameter assigned to multiple values")
+
+    if isinstance(creator, FunctionType):
```
alexjuda, line 451: Does this check work as expected if `creator` isn't a `def` or `lambda` but an arbitrary object with `__call__` method?

alexjuda (changes requested):

src/python/zquantum/core/utils.py
simonwa7: Nope! Fixed here: 96d45a1

src/python/zquantum/core/utils.py
simonwa7: I'm actually not sure what the desired behavior would be in this case... 
Playing around with it, it would not classify a class with `__call__` as a function:
```
>>> from types import FunctionType
>>> class MyClass():
...     def __init__(self):
...             pass
...     def __call__(self):
...             pass
... 
>>> isinstance(MyClass(), FunctionType)
False
```

alexjuda: Do I understand correctly that the point of this `if` is to be able to call `inspect.signature(creator)` later on on the object? If so, perhaps [EAFP](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/) with a localized `try`/`catch` with exceptions that can be raised by `inspect.signature` would be more robust against edge cases?

max-radin: One example I can think of where you have a callable class is where you use `AnsatzBasedCostFunction` for `cost_function_specs`. In which case you want `create_object` to instantiate the object, not create a partial function. So I guess this scenario would be handled correctly.

max-radin:
@simonwa7 could you elaborate on the issues this refactor is intended to help with? Were there some use cases where `create_function`/`create_object` were causing problems?

alexjuda (approved):

PR #270, opened 2021-04-15
minimal precommit for actions 

mbeach42:
Minimal pre-commit that its 100% compatible with GitHub Actions currently

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,87 +1,32 @@
 # See https://pre-commit.com for more information
-# See https://pre-commit.com/hooks.html for more hooks
-#
+# Minimal pre-commit to be compatible with current GitHub Actions style.yml
+
 default_language_version:
   python: python3.7
 
 default_stages: [commit, push]
 
 repos:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,87 +1,32 @@
...
-
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.9.0
     hooks:
       - id: flake8
-        additional_dependencies: [
-            flake8-comprehensions==3.2.2, # for better list comprehensions
-            flake8-docstrings==1.5.0,
-          ]
+        args: ["--ignore=E203,E266,F401", "--max-line-length=88"]
```
alexjuda, line 27: Shouldn't we use config from the `.flake8` file?

alexjuda (changes requested):

.pre-commit-config.yaml
mbeach42: okay, uploaded flake8 file to fix this

alexjuda (approved):
👍

PR #272, opened 2021-04-15
More typing fixes

max-radin:


src/python/zquantum/core/bitstring_distribution/distance_measures/mmd.py
```diff
@@ -51,19 +51,21 @@ def compute_rbf_kernel(x_i: np.ndarray, y_j: np.ndarray, sigma: float) -> np.nda
...
+    """Compute the multi-gaussian (RBF) kernel matrix K, with K_ij = 1/N * Sum_n [exp(-gamma_n |x_i - y_j|^2)] with n = 1,...,N and gamma = 1/(2*sigma).
 
-        Args:
-            x_i (np.array): Samples A (integers).
-            y_j (np.array): Samples B (integers).
-            sigmas (np.array): The list of bandwidths of the multi-gaussian kernel.
+    Args:
+        x_i (np.array): Samples A (integers).
+        y_j (np.array): Samples B (integers).
+        sigmas (np.array): The list of bandwidths of the multi-gaussian kernel.
```
mstechly, line 62: So the typing says `List` and docstring says `np.array`.
Actually, since you already have type in the definition, you can get rid of that in the docstring to avoid redundancy + it should be `np.ndarray` and not `np.array`.

src/python/zquantum/core/wip/circuits/_gates.py
```diff
@@ -13,6 +13,14 @@
 
 Parameter = Union[sympy.Symbol, Number]
 
+def _get_free_symbols(parameters: Tuple[Parameter, ...]) -> Iterable[sympy.Symbol]:
```
mstechly, line 16: Not sure about what this refactor actually does. Could you elaborate?

src/python/zquantum/core/wip/circuits/conversions/cirq_conversions.py
```diff
@@ -139,19 +139,22 @@ def _rotation(angle: Parameter) -> cirq.EigenGate:
 qubit_index = attrgetter("x")
 
 
-@overload
-def export_to_cirq(gate: _gates.Gate) -> cirq.Gate:
-    pass
+# The overloads are commented out because mypy does not yet support overloads
```
mstechly, line 142: I'm not sure if that's a good enough reason to remove these functions 🤔

src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py
```diff
@@ -303,7 +305,12 @@ def _import_controlled_qiskit_op(
     wrapped_qubits = qiskit_qubits[qiskit_gate.num_ctrl_qubits :]
     wrapped_op = _import_qiskit_op(qiskit_gate.base_gate, wrapped_qubits)
     qubit_indices = map(_import_qiskit_qubit, qiskit_qubits)
-    return wrapped_op.gate.controlled(qiskit_gate.num_ctrl_qubits)(*qubit_indices)
+    if isinstance(wrapped_op, _gates.GateOperation):
+        return wrapped_op.gate.controlled(qiskit_gate.num_ctrl_qubits)(*qubit_indices)
+    else:
+        raise NotImplementedError(
```
mstechly, line 311: I'm not sure about this as well.

src/python/zquantum/core/wip/circuits/conversions/cirq_conversions.py
max-radin: Overloads are only for type checking, they don't actually affect the behavior of the function. Though maybe there's a more elegant way though to tell mypy to ignore them besides commenting them out.

mstechly: I'm not sure if we don't have a clash between [this overload](https://www.python.org/dev/peps/pep-3124/) and [this overload](https://pypi.org/project/overload/) here.
Cause I'm pretty confident that I have used the latter.

src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py
max-radin: mypy wasn't happy with the original code because `wrapped_op` could be either be a `GateOperation` or a `AnonGateOperation`, but only `GateOperation` has a `gate` attribute. So I think the original behavior would have been to raise an `AttributeError` when `wrapped_op` is a `AnonGateOperation`, whereas now it raises `NotImplementedError`. @dexter2206 or @alexjuda what are your thoughts?

src/python/zquantum/core/wip/circuits/_gates.py
max-radin: So originally, `MatrixFactoryGate` was borrowing the `free_symbols` method from `Gate`: https://github.com/zapatacomputing/z-quantum-core/blob/6f034fa0353e2307cc2e0cfe55e8d343d777b51d/src/python/zquantum/core/wip/circuits/_gates.py#L247-L250

However if you tried to call the `free_symbols` property of a `MatrixFactoryGate`, mypy would complain because `self` argument had the wrong type. So I just put the logic for getting the symbols into the function `_get_free_symbols`, and refactored the methods of `Gate` and `MatrixFactoryGate` to call that function.

src/python/zquantum/core/bitstring_distribution/distance_measures/mmd.py
max-radin: Fixed in 4a5f84a92a6587f48d6cb7e0ecd7c56343ba28af

src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py
max-radin: Also note that I updated `_import_qiskit_op` to handle the `NotImplementedError`. @dexter2206 @alexjuda would be good if you could confirm whether this makes sense to do. https://github.com/zapatacomputing/z-quantum-core/blob/ce48d66e5b03ba10b0e29eb4da4466da4352e4b0/src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py#L260-L273

src/python/zquantum/core/wip/circuits/conversions/cirq_conversions.py
max-radin: Ohh that is confusing. The cirq conversions are importing `overload` from `typing` and not the `overloading` module... @dexter2206 can you confirm that this was what was intended?

src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py
alexjuda: You've found a bug! 🙂 Turns out I forgot about adding (and testing) support for converting controlled custom gates between our circuits and Qiskit. I'll be happy to fix it, but I'd rather do that in another PR so this one doesn't get convoluted. 

My suggestion would be to leave the `raise NotImplementedError(...)` here + abstain from catching the error around line 270 to let it fail explicitly until I fix it. Would that make sense?

src/python/zquantum/core/wip/circuits/_gates.py
mstechly: 👍

src/python/zquantum/core/wip/circuits/conversions/qiskit_conversions.py
max-radin: Makes sense! Removed the catching of `NotImplementedError` in 0bf1cfdbfc7ea1e83c71799e33a4d580a46d3fe8

src/python/zquantum/core/wip/circuits/conversions/cirq_conversions.py
alexjuda: Added a facade to overcome the mypy errors (01b236a). Mypy still reports some errors in `pyquil_conversions`, I'll look into it later

mstechly (approved):

PR #273, opened 2021-04-15
Added compatibility decorator to mock backends.

mstechly:


src/python/zquantum/core/interfaces/mock_objects.py
```diff
@@ -25,23 +29,25 @@ class MockQuantumBackend(QuantumBackend):
     def __init__(self, n_samples=None):
         super().__init__(n_samples)
 
-    def run_circuit_and_measure(self, circuit, n_samples=None, **kwargs):
+    @compatible_with_old_type(Circuit, new_circuit_from_old_circuit)
+    def run_circuit_and_measure(
+        self, circuit: NewCircuit, n_samples: int = None, **kwargs
```
dexter2206, line 34: I think `n_samples` should be types as `Optional[int]`, because of its default `None` value.

dexter2206 (approved):
One little fix recommended, otherwise looks good.

PR #277, opened 2021-04-20
Migrate backend interface

alexjuda:
Prepares the backend interface for migrating backend implementations to use new circuits.

(paired with @dexter2206)

### DIFF
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -11,7 +11,7 @@ from ..bitstring_distribution import (
     BitstringDistribution,
     create_bitstring_distribution_from_probability_distribution,
 )
-from ..circuit import Circuit, CircuitConnectivity
+from ..circuit import CircuitConnectivity
 from ..measurement import ExpectationValues, Measurements, expectation_values_to_real
 from ..openfermion import change_operator_type, get_expectation_value
 from ..wip.circuits._compatibility import AnyCircuit
@@ -192,7 +192,7 @@ class QuantumSimulator(QuantumBackend):
 
     @overrides
     def get_expectation_values(
-            self, circuit: AnyCircuit, operator: SymbolicOperator, **kwargs
+        self, circuit: AnyCircuit, operator: SymbolicOperator, **kwargs
     ) -> ExpectationValues:
         """Run a circuit and measure the expectation values of given operator.
 
--- a/src/python/zquantum/core/interfaces/backend.py
+++ b/src/python/zquantum/core/interfaces/backend.py
@@ -14,6 +14,7 @@ from ..bitstring_distribution import (
 from ..circuit import Circuit, CircuitConnectivity
 from ..measurement import ExpectationValues, Measurements, expectation_values_to_real
 from ..openfermion import change_operator_type, get_expectation_value
+from ..wip.circuits._compatibility import AnyCircuit
 
 
 class QuantumBackend(ABC):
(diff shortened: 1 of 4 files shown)


src/python/zquantum/core/interfaces/backend.py
```diff
@@ -64,7 +65,7 @@ def run_circuit_and_measure(
 
     def run_circuitset_and_measure(
```
mstechly, line 66: What about changing the name of this function as well, so we get rid of `circuitset` for good?

We could sth like this: 
```
def runcircuitset_and_measure(...):
    DepreciationWarning("runcircuitset_and_measure will now be callend XXX")
    return XXX(...)
```
It will also help with in case someone used `circuit_set` as a keyword argument instead of positional.
Not sure what the good name for XXX would be, as `run_circuits_and_measure` is too similar to `run_circuit_and_measure` and could cause confusion.
Perhaps `run_multiple_circuits_and_measure`.

mstechly (changes requested):
It also requires adding changes to the tests (for naming consistency sake) and steps.

src/python/zquantum/core/interfaces/backend.py
alexjuda: According to the docstrings the main use case for having methods with collections is batching. WDYT about specifying this explicitly in the name? Sth like:
```
def run_and_measure(
    self, circuit: AnyCircuit, n_samples: Optional[int] = None, **kwargs
):
    pass

def run_and_measure_batch(
    self, circuits_batch: Iterable[AnyCircuit], n_samples: Optional[int] = None, **kwargs
):
    pass

def get_expectation_values(
    self, circuit: AnyCircuit, operator: SymbolicOperator, **kwargs
) -> ExpectationValues:
    pass

def get_expectation_values_batch(
    self, circuits_batch: Iterable[AnyCircuit], operator: SymbolicOperator, **kwargs
) -> List[ExpectationValues]:
    pass

Comment (mstechly) [reply src/python/zquantum/core/interfaces/backend.py:66]: Yeah, I like using the `batch` word.
CC: @max-radin @simonwa7 – sounds good?

Comment (simonwa7) [reply src/python/zquantum/core/interfaces/backend.py:66]: Not sure if this is the time/place for this, but `run_circuit_and_measure` should also be deprecated as it's just an instance of running a batch of 1

Comment (simonwa7) [reply src/python/zquantum/core/interfaces/backend.py:66]: I would also be in favor of just: 
```
def run_and_measure(
    self, circuits: Iterable[AnyCircuit], n_samples: Optional[int] = None, **kwargs
):
    pass

def get_expectation_values(
    self, circuit:  Iterable[AnyCircuit], operator: SymbolicOperator, **kwargs
) -> ExpectationValues:
    pass
```

mstechly: I think @simonwa7 is right – both in how it should be handled and that it might not be time to do all that right now and we should get back to it later.

alexjuda:
@mstechly 

> It also requires adding changes to the tests (for naming consistency sake) and steps.

We've discussed this with @dexter2206:
1. Backend test base classes can be kept for checking backwards compatibility - we want backends to support old circuits for now. We'll add new tests with new circuits separately. Perhaps this will be a nice opportunity to go with the [contracts instead of inheritance](https://github.com/zapatacomputing/z-quantum-qaoa/pull/35).
2. Changing steps implementation to use new circuits would cause the code to pass new circuits to backends. This won't work until all backends are migrated.

This PR only extends the backend interface in a backwards-compatible way. My suggestion is to merge it like it is and introduce the above changes in subsequent PRs, after handling the backends.

src/python/zquantum/core/interfaces/backend.py
alexjuda: There are concerns against doing too much changes at once (see my comment below the PR). @dexter2206 and I would suggest to do the renaming on some other occasion.

mstechly (approved):
@alexjuda This is the way
✅

alexjuda:
I accidentally added pushed commits that tried to refactor the steps. I've reverted the one that changed the logic, the only changes left are docstring updates, so I guess it'd no harm to merge them since they're already here 🤷

PR #282, opened 2021-04-22
Typing fixes 4

max-radin:
This PR fixes the remaining typing issues in `src`.

### DIFF
--- a/src/python/zquantum/core/wip/cost_function.py
+++ b/src/python/zquantum/core/wip/cost_function.py
@@ -9,12 +9,12 @@ from .estimators.estimation import (
     estimate_expectation_values_by_averaging,
     evaluate_estimation_circuits,
 )
-from ..interfaces.functions import function_with_gradient, StoreArtifact
+from ..interfaces.functions import function_with_gradient, StoreArtifact, FunctionWithGradient, FunctionWithGradientStoringArtifacts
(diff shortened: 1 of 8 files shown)


src/python/zquantum/core/wip/cost_function.py
```diff
@@ -154,9 +153,9 @@ def sum_expectation_values(expectation_values: ExpectationValues) -> ValueEstima
     precision = None
 
     if expectation_values.estimator_covariances:
-        estimator_variance = 0
+        estimator_variance = 0.0
         for frame_covariance in expectation_values.estimator_covariances:
-            estimator_variance += np.sum(frame_covariance, (0, 1))
+            estimator_variance += float(np.sum(frame_covariance, (0, 1)))
```
mstechly, line 158: I'm not sure how this line makes sense.
Why do we sum over `frame_covariance` and `(0,1)`?

mstechly (approved):
Overall looks good but left a comment.

src/python/zquantum/core/wip/cost_function.py
max-radin: `frame_covariance` is a two-dimensional array where entry `i,j` is the covariance between term `i` and term `j`. To calculate the precision we have to sum over all pairs `i,j`, which corresponds to summing both over axis `0` and axis `1` of the array.

PR #288, opened 2021-04-29
177725377 migrate time evolution code

dexter2206:
This creates a wip package with time evolution code that produces new circuits. Also, the tests are cleaned up and parameterized.

### DIFF
--- a/src/python/zquantum/core/wip/evolution.py
+++ b/src/python/zquantum/core/wip/evolution.py
@@ -44,7 +44,7 @@ def time_evolution(
 def _adjust_gate_angle(operation: circuits.GateOperation, time):
     """Adjust angle in gate operation to account for evolution in time.
 
-    Since this handles outputs from pyuil.paulis.exponentiate, the only
+    Since this handles outputs from `pyquil.paulis.exponentiate`, the only
     time-dependent gates are RZ and PHASE (other rotations have fixed
     angles as they correspond to change of basis).
 
--- a/src/python/zquantum/core/wip/evolution.py
+++ b/src/python/zquantum/core/wip/evolution.py
@@ -1,9 +1,11 @@
-from functools import reduce
 import operator
+from functools import reduce
+from typing import Union, Tuple, List
+
 import numpy as np
+import pyquil.paulis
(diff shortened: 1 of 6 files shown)


src/python/zquantum/core/wip/evolution.py
```diff
@@ -0,0 +1,192 @@
...
+                        repeated_circuit, different_circuit, trotter_order, position
+                    )
+                )
+                final_factors.append(factor)
+        return output_circuits, final_factors
+    else:
+        return single_trotter_derivatives, output_factors
+
+
+def generate_circuit_sequence(repeated_circuit, different_circuit, length, position):
```
mstechly, line 165: Shouldn't this be private?

mstechly: Also, lacks typing.

tests/zquantum/core/wip/evolution_test.py
```diff
@@ -0,0 +1,200 @@
...
+    generate_circuit_sequence,
+    time_evolution,
+    time_evolution_derivatives,
+    time_evolution_for_term,
+)
+
+PAULI_STRING_TO_CIRQ_GATE = {"XX": cirq.XX, "YY": cirq.YY, "ZZ": cirq.ZZ}
+
+
+def _cirq_exponentiate_term(term, qubits, time, order):
```
mstechly, line 19: I would change `order` to `trotter_order` so it's more explicit.

mstechly (approved):
Overall looks good, with minor comments.
Good to merge after fixing.

src/python/zquantum/core/wip/evolution.py
dexter2206: I left it as a public function because it was public in the old `evolution.py`. However, since no one i using this wip package, it might indeed be a good idea to make it private. This and the lack of typing have been addressed in 432aa86.

tests/zquantum/core/wip/evolution_test.py
dexter2206: Addressed in 8974b5a

PR #295, opened 2021-05-04
173183192 more flexible graphs

martamau:
I think we should merge this branch into dev (and master then) as this is what we were using for FLIP simulations. I haven't worked on this branch but maybe the contributors could help taking a look at this!


src/setup.py
```diff
@@ -18,6 +18,7 @@
         "pytest>=5.3.5",
         "networkx==2.4",
         "numpy>=1.18.1",
+        "sympy==1.5",
```
mstechly, line 21: Sympy is already here, with 1.7 in line 23.

martamau: I think this pinned version was set because of tensorflow/tensorflow quantum compatibility. Agree that we should stick to the version in dev/master though

alexjuda:
The discussion moved to #362.

PR #300, opened 2021-05-06
Fix handling inactive qubits past last active qubit

dexter2206:
Currently, converting old circuit to new (wip) circuit discards inactive qubits past the last active qubit. This is because we used old -> Cirq -> new conversion strategy, and already old -> Cirq part loses those inactive qubits.

This PR fixes this by checking the old circuit for the qubit with the largest index and then explicitly setting `n_qubits` in the new circuit.

### DIFF
--- a/src/python/zquantum/core/wip/circuits/_compatibility.py
+++ b/src/python/zquantum/core/wip/circuits/_compatibility.py
@@ -9,4 +9,9 @@ AnyCircuit = Union[OldCircuit, NewCircuit]
 
 
 def new_circuit_from_old_circuit(old: OldCircuit) -> NewCircuit:
-    return import_from_cirq(old.to_cirq())
+    if not old.qubits:
+        return NewCircuit([])
+
+    pre_new_circuit = import_from_cirq(old.to_cirq())
+    n_qubits = max([qubit.index for qubit in old.qubits]) + 1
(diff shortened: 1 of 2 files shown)


src/python/zquantum/core/wip/circuits/_compatibility.py
```diff
@@ -9,4 +9,9 @@
 
 
 def new_circuit_from_old_circuit(old: OldCircuit) -> NewCircuit:
-    return import_from_cirq(old.to_cirq())
+    if not old.qubits:
+        return NewCircuit([])
+
+    pre_new_circuit = import_from_cirq(old.to_cirq())
+    n_qubits = max([qubit.index for qubit in old.qubits]) + 1
+    return NewCircuit(pre_new_circuit.operations, n_qubits=n_qubits)
```
alexjuda, line 17: Will it work as expected if `old` contains any custom gates? The former `import_from_cirq(old.to_cirq())` one-liner would capture that, assuming we support importing custom gates from Cirq

alexjuda: Ahh, nevermind. I forgot we don't pass custom gate definitions via `__init__` anymore.

alexjuda (approved):

PR #304, opened 2021-05-12
Adding support for converting U3 to and from cirq

max-radin:


tests/zquantum/core/circuit/_gate_test.py
```diff
@@ -22,6 +25,8 @@ def create_gate(self, gate_name, qubit_indices=[0, 1], params=None):
             params = []
             if gate_name in self.one_parameter_gates:
                 params = [1.0]
+            if gate_name in self.three_parameter_gates:
+                params = [1.0, 1.0, 1.0]
```
simonwa7, line 29: It would be good to make these different values (e.g. `[1.0, 2.0, 3.0]`) so that an incorrect assignment/ordering of the parameter values would be caught.

max-radin: Fixed in 5589124890e374e91238354bc27b01a968af6129

simonwa7 (approved):

PR #305, opened 2021-05-12
Fix flake8 errors in wip module

alexjuda:


src/python/zquantum/core/wip/circuits/_builtin_gates.py
```diff
@@ -33,7 +33,9 @@ def builtin_gate_by_name(name) -> GateRef:
 Y = _gates.MatrixFactoryGate("Y", _matrices.y_matrix, (), 1, is_hermitian=True)
 Z = _gates.MatrixFactoryGate("Z", _matrices.z_matrix, (), 1, is_hermitian=True)
 H = _gates.MatrixFactoryGate("H", _matrices.h_matrix, (), 1, is_hermitian=True)
-I = _gates.MatrixFactoryGate("I", _matrices.i_matrix, (), 1, is_hermitian=True)
+I = _gates.MatrixFactoryGate(  # noqa: E741
```
mstechly, line 36: Why this particular line? As it's basically the same as all the previous gates?

mstechly (approved):
I would make some of the line breaks in slightly different places, but overall looks good.
There's just this one comment where I don't understand why change was required.

src/python/zquantum/core/wip/circuits/_builtin_gates.py
alexjuda: This was the only line that flake8 complained about - the error was that `I` is an ambiguous name. It doesn't mind `Z` or `H` 🤷

PR #314, opened 2021-05-13
Updated estimation to new circuits.

mstechly:
Notes:
1. Tests will fail because other code depends on estimators – however, I don't want to make a monstrous PR with everything, so we'll figure out tests later
2. There are some formatting errors I got, will fix them later.

### DIFF
--- a/tests/zquantum/core/estimation/estimation_test.py
+++ b/tests/zquantum/core/estimation/estimation_test.py
@@ -1,55 +1,26 @@
+from functools import partial
+
 import numpy as np
 import pytest
 import sympy
 from openfermion import IsingOperator, QubitOperator, qubit_operator_sparse
-from pyquil import Program
-from pyquil.gates import RX, RY, RZ, X, Y
-from functools import partial
-from zquantum.core import estimation
-from zquantum.core.circuit import Circuit, Qubit, Gate
-from zquantum.core.interfaces.mock_objects import (
-    MockQuantumBackend,
-    MockQuantumSimulator,
-)
+from zquantum.core.estimation import (allocate_shots_proportionally,
+                                      allocate_shots_uniformly,
+                                      calculate_exact_expectation_values,
+                                      estimate_expectation_values_by_averaging,
+                                      evaluate_estimation_circuits,
+                                      get_context_selection_circuit_for_group,
+                                      group_greedily, group_individually,
+                                      perform_context_selection)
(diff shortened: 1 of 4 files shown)


tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -1,55 +1,26 @@
...
-from pyquil import Program
-from pyquil.gates import RX, RY, RZ, X, Y
-from functools import partial
-from zquantum.core import estimation
-from zquantum.core.circuit import Circuit, Qubit, Gate
-from zquantum.core.interfaces.mock_objects import (
-    MockQuantumBackend,
-    MockQuantumSimulator,
-)
+from zquantum.core.estimation import (allocate_shots_proportionally,
```
alexjuda, line 7: What kind of formatter did you use? Running `black` on my side resulted in newlines between the `(` and `)`

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -1,55 +1,26 @@
+from functools import partial
+
```
alexjuda, line 2: (This applies to the tests in general.) You mentioned some of the tests are expected to fail. WDYT about marking these tests with [skip](https://docs.pytest.org/en/latest/how-to/skipping.html) in a single, separate commit? This way we should be able to easily re-enable them when we're ready.

alexjuda (changes requested):

tests/zquantum/core/estimation/estimation_test.py
mstechly: This was my first commit with pre-commit hooks configured, perhaps something is wrong with my configuration –will be investigating later.

tests/zquantum/core/estimation/estimation_test.py
mstechly: So adding sth like:
`@pytest.mark.skip(reason="These tests will fail until CostFunction is migrated to new circuits")`
?

mstechly: Just wondering – is it better to use `skip` or `xfail` in such case?

alexjuda: I'd use `skip` because it doesn't actually run the tests, saving some time + it provides a natural place to add a commentary

alexjuda (approved):

PR #317, opened 2021-05-17
177727859 fix flake8 vol4

max-radin:
Note that for `circuit/__init__.py`, I put a flag to ignore F403 (`*` imports). The circuit module is going to be replaced soon so it doesn't seem worth fixing this right now.


src/python/zquantum/core/bitstring_distribution/distance_measures/__init__.py
```diff
@@ -1,2 +1,2 @@
-from .clipped_negative_log_likelihood import *
-from .mmd import *
+from .clipped_negative_log_likelihood import compute_clipped_negative_log_likelihood
+from .mmd import compute_mmd, compute_multi_rbf_kernel, compute_rbf_kernel
```
alexjuda, line 2: Good catch! It's not always feasible to replace the wildcard with explicit imports, but here it totally makes sense.

alexjuda (approved):

PR #321, opened 2021-05-18
173183192 more flexible graphs refactor

simonwa7:


simonwa7:
@martamau this is a small recommendation to refactor this code a bit to make it clearer. The basic idea is to rename the `random_weights` argument to `weights` and no longer allow it to be a `bool`, but assume it is a `string`. This will reduce some confusion about what the input types should be. Without knowing more about how this is used, I think this is a nice ccompromise between a fully-fledged refactor.

alexjuda:
Discussion moved to #362

PR #325, opened 2021-05-19
Skipping circuits for evaluating the constant term

jgonthier:
This modifies `estimate_expectation_values_by_averaging` so that if an EstimationTask only contains the constant term, it is not submitted to the backend as we already know the resulting expectation value.

In addition, if an EstimationTask has 0 shot and non-constant terms, we raise a RuntimeError as I don't really know how to handle this. In the future we may consider setting expectation values to 0 and variances to infinity or their maximum value. (Then we'd have to consider the case where the EstimationTask also contains a constant term, in which case that expectation value should be 1.0 and the variances 0.0)


tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -374,13 +539,15 @@ def test_calculate_exact_expectation_values(self, simulator, estimation_tasks):
         )
         assert len(expectation_values_list) == 3
         for expectation_values, task in zip(expectation_values_list, estimation_tasks):
+            if () in task.operator.terms.keys():
+                for i, term in enumerate(task.operator.terms.items()):
+                    if term[0] == ():
+                        assert expectation_values.values[i] == term[1]
```
aakunitsa, line 545: I fixed the `MockQuantumSimulator` since it always stored 1. in the expectation values, causing this test to fail.

aakunitsa (approved):
I had to fix the mock simulator to make all the tests pass but otherwise, everything looks great!

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -219,6 +219,57 @@ def evaluate_estimation_circuits(
     ]
 
 
+def get_estimation_tasks_to_measure(
```
mstechly, line 222: I think the way you do it is ok, but I think we could improve it slightly:
1. Take all the estimation tasks
2. Split it into two groups: executable and non-executable
3. Then treat "executable" as usual and "non-executable" separately
4. Combine the results at the end.

I think it still would require keeping track of the indices in the expectation values, but at least it would more explicit.
Also, `get_estimation_tasks_to_measure` actually does two things – it gives you the estimation tasks you want to use and it evaluates those with constant terms. And therefore it breaks [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

 (which might be cleaner and more in line with the conventions we follow).

jgonthier: I think your suggestion makes a lot of sense ! I'm thinking that I will then split `get_estimation_tasks_to_measure` into two functions:
- one that will split the EstimationTasks into two groups
- one that evaluates EstimationTasks that have constant terms

Does that seem like a good plan to you?

mstechly: Yes, sounds like plan!

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
     ]
 
 
+def split_constant_estimation_tasks(
+    estimation_tasks: List[EstimationTask],
+) -> Tuple[List[EstimationTask], List[EstimationTask], List[int], List[int]]:
+    """This function splits a give list of EstimationTask into two: one that
```
mstechly, line 231:
```suggestion
    """This function splits a given list of EstimationTask into two: one that
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+    contains EstimationTasks that only contain constants, and one that contains
+    EstimationTasks that should actually be submitted to a backend. EstimationTask
```
mstechly, line 233:
```suggestion
    EstimationTasks that have non-constant terms as well.
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+    that require 0 shot generate an error.
+
+    Args:
+        estimation_tasks (List[EstimationTask]): The list of estimation tasks for which
```
mstechly, line 237:
```suggestion
        estimation_tasks: The list of estimation tasks for which
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+                         Expectation Values are wanted.
+
+    Returns:
+        estimation_tasks_to_measure (List[Circuit]): A new list of estimation tasks that only
```
mstechly, line 241:
```suggestion
        estimation_tasks_to_measure: A new list of estimation tasks that only
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+                                     contains the ones that should actually be submitted to the
+                                     backend
+        estimation_tasks_for_constants (List[Circuit]): A new list of estimation tasks that
```
mstechly, line 244:
```suggestion
        estimation_tasks_for_constants: A new list of estimation tasks that
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+                                        contains the EstimationTasks with only constant terms
+        indices_to_measure (List[int]): A list containing the indices of the EstimationTasks we will
```
mstechly, line 246:
```suggestion
        indices_to_measure: A list containing the indices of the EstimationTasks we will
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+                            actually measure, i.e. the ith estimation_tasks_to_measure expectation
+                            value will go into the indices_to_measure[i] position.
+        indices_for_constants (List[int]): A list containing the indices of the EstimationTasks for
```
mstechly, line 249:
```suggestion
        indices_for_constants: A list containing the indices of the EstimationTasks for
```

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+                               constant terms.
+    """
+
+    estimation_tasks_to_measure = []
+    estimation_tasks_for_constants = []
+    indices_to_measure = []
+    indices_for_constants = []
+    for i, e in enumerate(estimation_tasks):
```
mstechly, line 257: I would change it to `i, task` ;)

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,6 +225,88 @@ def evaluate_estimation_circuits(
...
+        estimation_tasks (List[EstimationTask]): The list of estimation tasks for which
+                         Expectation Values are wanted, they must only contain constant terms.
+
+    Returns:
+        expectation_values (List[ExpectationValues]): the expectation values over constant terms,
+                            with their correlations and estimator_covariances.
+    """
+
+    expectation_values = []
+    for e in estimation_tasks:
```
mstechly, line 292: Please change `e` to `task`.

mstechly (approved):

src/python/zquantum/core/estimation/_estimation.py
jgonthier: Addressed in be2a71d78c59d2ad544bc9a23dacd3f5a7dad106

src/python/zquantum/core/estimation/_estimation.py
jgonthier: Addressed in be2a71d78c59d2ad544bc9a23dacd3f5a7dad106

PR #336, opened 2021-05-26
Migrate optimize step to new circuits

alexjuda:


tests/acceptance_tests/optimize_test.py
```diff
@@ -8,158 +9,164 @@
 TARGET_OPERATOR = QubitOperator("X0 X1 Z2 Y4", 1.5)
 
 
+# To regenerate the circuit, run:
+# import zquantum.core.wip.circuits as new_circuits
```
dexter2206, line 13: Let's also add `sympy` import here to make the snippet self-contained.

dexter2206 (changes requested):

tests/acceptance_tests/optimize_test.py
alexjuda: fixed in 4c14120

dexter2206 (approved):

PR #347, opened 2021-05-31
Generate list of ordered bitstrings of given lenght

martamau:
I use this function very often and need to copy pasta it in all my scripts so I figured it would be useful to include it in our utils :)


src/python/zquantum/core/utils.py
```diff
@@ -647,3 +647,23 @@ def load_from_specs(specs: Specs):
     if isinstance(specs, str):
         specs = json.loads(specs)
     return create_object(specs)  # type: ignore
+
+
+def get_ordered_list_of_bitstrings(num_qubits: int):
```
mstechly, line 652: I'd move it to `measurements.py` instead of keeping it in `utils.py`

mstechly, line 652: Please add typing and then remove "Returns" from the docstring.
```suggestion
def get_ordered_list_of_bitstrings(num_qubits: int) -> List[str]:
```

src/python/zquantum/core/utils.py
```diff
@@ -647,3 +647,23 @@ def load_from_specs(specs: Specs):
...
+    """Create list of binary strings corresponding to 2^num_qubits integers
+    and save them in ascending order.
+
+    Args:
+        num_qubits: number of binary digits in each bitstring
+
+    Returns:
+        List[string]: the ordered bitstring representations of the integers
+    """
+    # Determine occupancy of each spin orbital
```
mstechly, line 662: I think this line is not necessary :)
```suggestion
```

tests/zquantum/core/utils_test.py
```diff
@@ -473,3 +474,9 @@ def test_scale_and_discretize(values, total, expected_result):
 def test_hf_rdm_energy(hamiltonian, ref_energy, nalpha):
     rdm = hf_rdm(nalpha, 1, 2)
     assert np.isclose(ref_energy, rdm.expectation(hamiltonian))
+
+
+@pytest.mark.parametrize("num_qubits", [2, 3, 5, 10])
+def test_ordered_bitstring(num_qubits):
+    bitstrings = get_ordered_list_of_bitstrings(num_qubits)
+    assert np.all([len(bit) == num_qubits for bit in bitstrings])
```
mstechly, line 482: 1. This test just checks whether the length of the bitstrings is correct, but it doesn't test whether it actually returned the correct bitstrings or in the correct order.
2. It should be `for bitstring in bitstrings` and not `for bit in bitstrings`.

mstechly (changes requested):

src/python/zquantum/core/utils.py
martamau: Why do you think this function is related to measurements? It simply generates orderder binary representation of integers, I put it here as we also have dec2bin and bin2dec here which are similar kind of utils

mstechly: Hmm 🤔 
Everything that has bitstrings looks like measurements to me, but I think you're right :)

mstechly (approved):

PR #351, opened 2021-06-03
Updated optimizer interface.

mstechly:


src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -13,39 +14,43 @@ class Optimizer(ABC):
...
-            'optimizers from continuing to declare "options" within their individual '
-            "constructors.",
-            DeprecationWarning,
-        )
-        if options is None:
-            options = {}
-        self.options = options
-        if "keep_value_history" not in self.options.keys():
-            self.options["keep_value_history"] = False
+    def __init__(self, recorder: _recorder) -> None:
```
simonwa7, line 21: Need to add typing. @dexter2206

simonwa7 (approved):

src/python/zquantum/core/interfaces/optimizer.py
dexter2206: I think in this case, we can't use `_recorder` as a type hint because it is just a function. It is rather hard to come up with a decent hint here, because ideally it would have to satisfy the following conditions:

- (must have) it is a function that accepts a function and returns some kind of the recorder
- it transforms function with gradient into the recorder with gradient
- it transforms function storing artifacts into recorder storing artifacts

I don't know if it is possible to satisfy all the constraints, but if we can satisfy the first one it should be possible to work around other two by using casts. Therefore, my initial recommendation would be to use something along the lines of:

```
AnyRecorder = Union[SimpleRecorder, ArtifactRecorder]
RecorderFactory = Callable[[Callable], AnyRecorder]
```
and then use `RecorderFactory` as a type of `recorder` argument. I expect MyPy to complain, but then we'll see what fixes need to be made.

PR #352, opened 2021-06-07
Fix running tests on ms windows

dexter2206:
This solves the problem of some serialization tests not working on MS Windows. The following changes were needed:

- Creating temporary files with `TemporaryDirectory` + `mkstemp` instead of `NamedTemporaryFile`.
- Explicit encoding of text files using `utf-8` (on Windows platform the default encoding can't encode some characters used in tests). This change requires adding the `encoding` parameter to `ensure_open`.

### DIFF
--- a/tests/zquantum/core/serialization_test.py
+++ b/tests/zquantum/core/serialization_test.py
@@ -258,8 +258,10 @@ def test_load_optimization_results_successfully_loads_optimization_result_from_f
 
 @pytest.fixture
 def tmp_path():
-    with tempfile.NamedTemporaryFile() as tmp_file:
-        yield tmp_file.name
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        handle, path = tempfile.mkstemp(dir=tmp_dir)
+        os.close(handle)  # This is needed, so that we can reopen on MS Windows
+        yield path
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


alexjuda (approved):

tests/zquantum/core/serialization_test.py
```diff
@@ -258,8 +258,10 @@ def test_load_optimization_results_successfully_loads_optimization_result_from_f
 
 @pytest.fixture
 def tmp_path():
-    with tempfile.NamedTemporaryFile() as tmp_file:
-        yield tmp_file.name
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        handle, path = tempfile.mkstemp(dir=tmp_dir)
```
alexjuda, line 262: What's the use of `mkstemp` in addition to `TemporaryDirectory` necessary? I think the latter one creates and deletes the temporary directory automatically so we could just generate a unique path and yield it. Did you consider it?

dexter2206: The `mkstemp` is used for creating file with unique name - it is guaranteed that two threads/processes won't try to create the same file. Arguably, this is not needed because 1) I don't suppose we want to run the same test in multiple threads 2) this can be replaced by using filename generated using e.g. current timestamp. I just thought it simpler to use `mkstemp` instead of reinventing the wheel 🙂

PR #355, opened 2021-06-07
Updating cirq conversions to give sympy matrix for custom gates

max-radin:
This PR fixes the issue @jgonthier encountered where gates imported form cirq as custom gates would raise an exception when constructing the unitary.

In a ideal world this would have been caught by static type checking. I think it was missed however because mypy does not provide type hints.


dexter2206 (changes requested):
Looks good. However, I recommend adding some test case that checks that the problem is indeed fixed.

mstechly (approved):

dexter2206 (approved):

PR #357, opened 2021-06-08
178220011 add free symbols

dexter2206:
This adds free_symbols to generic Operation Protocol, and also adds suitable implementation in every known implementer of Operation.


src/python/zquantum/core/circuits/_gates.py
```diff
@@ -1,28 +1,15 @@
...
 from dataclasses import dataclass, replace
-from functools import singledispatch
-from numbers import Number
 from typing import Callable, Dict, Iterable, Tuple, Union
 
 import numpy as np
 import sympy
 from typing_extensions import Protocol, runtime_checkable
 
+from ._operations import _get_free_symbols, Parameter, _sub_symbols
```
alexjuda: `_get_free_symbols` and `_sub_symbols` are no longer fileprivate, should the leading underscore be removed?

src/python/zquantum/core/circuits/_operations.py
```diff
@@ -0,0 +1,83 @@
...
+T = TypeVar("T", bound="Operation")
+
+
+class Operation(Protocol):
+    """Represents arbitrary operation applicable to a circuit or wavefunction."""
+
+    @property
+    def params(self) -> Tuple[Parameter, ...]:
+        """Parameters of this operation."""
+        raise NotImplementedError()
```
alexjuda, line 19: WDYT about decorating the protocol's methods with `@abstractmethod`? This would allow opt-in check for implementers. We could also remove the `raise NotImplementedError` lines, making codecov happy at the same time.

alexjuda (changes requested):

alexjuda (approved):

PR #362, opened 2021-06-10
Flexible graphs, vol 3

alexjuda:
This PR contains a refactor of a refactor (#321) of a refactor (#295) of graph generators.

# Comparison with #321

- using Python generators for distribution sampling instead of the string-based distribution description language. Check out tests or `steps/graph` for usage examples. I've added "samplers" for each of the distribution type I've seen in #321.
- deprecation of `generate_graph_from_specs`. I wanted to write a docstring that would document all the options and possible values for specs dict. It's intense even for the functionality already on `dev`, when I wanted to add description of options needed for the distribution descriptions I quit this idea altogether. I believe this level of dictionary nesting would be too deep to be useful, if there are use-cases where this could be handy we should try designing the graph generator API in a different way.

As a side note, I think that my proposed solution for samplers is similar to [the strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern). If we want to refine the idea further I believe this is the pattern to look for inspiration.

# Comparison with dev

- **[breaking change]** `generate_*_graph` functions in `zquantum.core.graph` used to accept boolean `random_weights` flag, now they accept a float generator. This breaking change was originally introduced in #321 and #295, I've just replaced string with generator
- graph generators can now use samplers to change the way they generate edge weights


src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+    while True:
+        yield random.uniform(min_value, max_value)
+
+
+def constant_sampler(value) -> Sampler:
+    return itertools.repeat(value)
+
+
+def static_sampler() -> Sampler:
+    return constant_sampler(1.0)
```
simonwa7, line 82: I would perhaps move to get rid of `static_sampler` in favor of just `constant_sampler`.

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+
+def choice_sampler(choices) -> Sampler:
+    while True:
+        yield random.choice(choices)
+
+
+def uniform_range_sampler(*range_args) -> Sampler:
+    while True:
+        choices = range(*range_args)
+        yield random.choice(choices)
```
simonwa7, line 93: Additionally, I don't find this one too intuitive. I would probably just use `choice_sampler` myself, but I'm not against keeping this if you think it has value.

steps/graph.py
```diff
@@ -1,27 +1,26 @@
...
+
+import zquantum.core.graph
 from zquantum.core.graph import save_graph
 
 
+def _make_sampler(random_weights: bool = False):
+    if random_weights:
+        return zquantum.core.graph.uniform_sampler()
+    else:
+        return zquantum.core.graph.static_sampler()
```
simonwa7, line 12: I noticed that you elect to use this function and `random_weights` as opposed to using `create_object` along with `sampler_specs` in order to generate the samplers from the workflow steps. Is there a reason for this because it seems like it wouldn't be possible to generate certain samplers such as `choice_sampler` or `uniform_sampler` with different `min_val` and `max_val`.

steps/graph.py
```diff
@@ -1,27 +1,26 @@
...
+
 def generate_random_graph_erdos_renyi(
     number_of_nodes: int,
     edge_probability: float,
     random_weights: bool = False,
     seed: Optional[int] = None,
 ):
-    graph = _generate_random_graph_erdos_renyi(
-        number_of_nodes, edge_probability, random_weights, seed
+    sampler = _make_sampler(random_weights)
```
simonwa7, line 21: I think what would look better would be:

```
sampler = create_object(sampler_specs)()
```

src/python/zquantum/core/graph.py
```diff
@@ -1,43 +1,42 @@
...
-    graph_dict = nx.readwrite.json_graph.node_link_data(graph)
-    graph_dict["schema"] = SCHEMA_VERSION + "-graph"
-    json.dump(graph_dict, f, indent=2)
-    f.close()
+    with ensure_open(filename, "w") as f:
+        graph_dict = {
+            **nx.readwrite.json_graph.node_link_data(graph),
+            "schema": GRAPH_SCHEMA,
+        }
+        json.dump(graph_dict, f, indent=2)
```
martamau, line 28: Shall we keep the `f.close()` line?

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
     return nodes_dict
 
 
+Sampler = Iterator[float]
+
+
+def uniform_sampler(min_value=0, max_value=1) -> Sampler:
```
martamau, line 72:
```suggestion
def uniform_sampler(min_value: float=0.0, max_value: float=1.0) -> Sampler:
```

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+    while True:
+        yield random.uniform(min_value, max_value)
+
+
+def constant_sampler(value) -> Sampler:
```
martamau, line 77:
```suggestion
def constant_sampler(value: float) -> Sampler:
```
Maybe we should we make sure this value is in [0,1].

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+    return itertools.repeat(value)
+
+
+def static_sampler() -> Sampler:
+    return constant_sampler(1.0)
+
+
+def choice_sampler(choices) -> Sampler:
```
martamau, line 85:
```suggestion
def choice_sampler(choices: np.ndarray) -> Sampler:
```

src/python/zquantum/core/graph.py
martamau: Maybe it was included in Fred's code for some reason?

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+        yield random.choice(choices)
+
+
+def uniform_range_sampler(*range_args) -> Sampler:
+    while True:
+        choices = range(*range_args)
+        yield random.choice(choices)
+
+
+def normal_sampler(mu, sigma) -> Sampler:
```
martamau, line 96:
```suggestion
def normal_sampler(mu: float = 0.0, sigma: float = 1.0) -> Sampler:
```

src/python/zquantum/core/graph.py
```diff
@@ -73,163 +66,223 @@ def generate_graph_node_dict(graph: nx.Graph) -> dict:
...
+        weight_sampler = static_sampler()
+
+    seed = graph_specs.get("seed")
+
     if type_graph == "erdos_renyi":
-        probability = graph_specs["probability"]
         graph = generate_random_graph_erdos_renyi(
-            num_nodes, probability, random_weights, seed
+            num_nodes,
+            graph_specs["probability"],
```
martamau, line 254: Just double checking - should this be "edge_probability"?

src/python/zquantum/core/graph.py
alexjuda: Not really – the `open` context manager (wrapped by `ensure_open`) closes the file on exiting the block 🙂

src/python/zquantum/core/graph.py
alexjuda: This sampler just produces `value` again and again, this doesn't have to be a probability, so I'd leave it unbounded.

src/python/zquantum/core/graph.py
alexjuda: Will do!

src/python/zquantum/core/graph.py
alexjuda: It doesn't have to be a numpy array, I'll provide the type hint in a sec

src/python/zquantum/core/graph.py
alexjuda: Yeah, it wasn't for me either, I'll remove it

src/python/zquantum/core/graph.py
alexjuda: yeah, makes more sense to me to, but I'd keep it for backwards compatibility. The whole function will be deprecated anyway

steps/graph.py
alexjuda: Updated in cde231f...f9d9601

simonwa7 (approved):

PR #363, opened 2021-06-10
Implement attribute forwarding in recorder

dexter2206:


mstechly (commented):
could you add the same to `FunctionWithGradient` and `FunctionWithGradientStoringArtifacts` ?
I'll test it then with my code and if that works would approve :)

mstechly (approved):

PR #366, opened 2021-06-14
Adding test and fixing error for sample elimination bug

jgonthier:
When generating bitstrings from measurements, if too many bitstrings are generated because of rounding, we eliminate the additional ones at random. This PR ensures we're not trying to eliminate bitstrings that are not in the original distribution.


src/python/zquantum/core/measurement.py
```diff
@@ -453,6 +453,50 @@ def get_expectation_value_from_frequencies(
...
+                     be removed from bitstring_samples
+    """
+    bitstring_counts = Counter(bitstring_samples)
+    new_samples = None
+    correct_samples = samples
+    for sample in samples:
+        bitstring = tuple([int(measurement_value) for measurement_value in sample])
+        if samples[sample] > bitstring_counts[bitstring]:
+            nresamples = samples[sample] - bitstring_counts[bitstring]
+            samples[sample] = bitstring_counts[bitstring]
```
alexjuda, line 482: WDYT about copying `samples` before the `for` loop so that this line doesn't mutate the argument?

src/python/zquantum/core/measurement.py
```diff
@@ -453,6 +453,50 @@ def get_expectation_value_from_frequencies(
...
+            distribution_dict = leftover_distribution.distribution_dict
```
alexjuda, line 483: Similarly,
```suggestion
            distribution_dict = dict(leftover_distribution.distribution_dict)
```

tests/zquantum/core/measurement_test.py
```diff
@@ -857,3 +858,81 @@ def test_get_measurements_representing_distribution_gives_exactly_right_counts(
...
+                    "1010": 0.008,
+                    "1110": 0.008,
+                    "1101": 0.008,
+                    "0110": 0.9102,
+                }
+            )
+        ],
+    )
+    def test_get_measurements_representing_distribution_does_not_remove_bitstrings_not_sampled(
+        self, bitstring_distribution
```
alexjuda, line 885: These lines caused stylechecks to [fail](https://github.com/zapatacomputing/z-quantum-core/runs/2823555534?check_suite_focus=true). The runner prints the failed command that you can invoke locally to replicate the issue:
```
flake8 --ignore=E203,E266,F401,W503 --max-line-length=88 src/python/zquantum/core steps tests/zquantum/core tests/acceptance_tests
```

tests/zquantum/core/measurement_test.py
```diff
@@ -857,3 +858,81 @@ def test_get_measurements_representing_distribution_gives_exactly_right_counts(
...
+    ):
+        number_of_samples = 100
+        max_number_of_trials = 100
+        cnt = 0
+        for _ in range(max_number_of_trials):
+            measurements = Measurements.get_measurements_representing_distribution(
+                bitstring_distribution, number_of_samples
+            )
```
alexjuda, line 893: At the moment this test only verifies that no exception is raised. Is there any other behavior to assert upon?

src/python/zquantum/core/measurement.py
```diff
@@ -453,6 +453,50 @@ def get_expectation_value_from_frequencies(
...
+            distribution_dict[sample] = 0
+            leftover_distribution = BitstringDistribution(distribution_dict, True)
+            new_samples = sample_from_probability_distribution(
```
alexjuda, line 486: Could you help me understand why `new_samples` having any elements is the condition for recursion? 

Also, this might be easily changed to a `while` loop (in principle any tail recursion can).

alexjuda (changes requested):

src/python/zquantum/core/measurement.py
jgonthier: Fixed in 2a6f36d8f784948f037f5152374e935f1ec99216

src/python/zquantum/core/measurement.py
jgonthier: Fixed in 2a6f36d8f784948f037f5152374e935f1ec99216

tests/zquantum/core/measurement_test.py
jgonthier: No, that is all that is checked here. The reason for adding this test is that before the bugfix, it would raise an Exception for trying to remove a non-existing element from a list. In case further modification of the code reintroduce this bug, it would be caught here.

src/python/zquantum/core/measurement.py
jgonthier: Fixed in 2a6f36d8f784948f037f5152374e935f1ec99216

tests/zquantum/core/measurement_test.py
jgonthier: Fixed in c8135ca9b0a924a1aa8669e943bee4a7fc34fd19

tests/zquantum/core/measurement_test.py
alexjuda: would it make sense to rename the test case to reflect that, e.g. `test_get_measurements_representing_distribution_doesnt_raise`?  This would remove the need for the comment :)

alexjuda (approved):
Added a minor test rename suggestion, but otherwise LGTM

tests/zquantum/core/measurement_test.py
jgonthier: Good point! Done in 005181dd6e9aa6c054525a3aff8dd0168bf48235

PR #375, opened 2021-06-22
178433179 implement xpowgate and ypowgate decomposition

dexter2206:
This PR implements gate decomposition engine. The engine, designed with @alexjuda , is inspired by formal grammars. Decomposition of gates (or more precisely, circuit operations) is defined via objects implementing `DecompositionRule`. Each such rule comprises:

- `predicate` determining whether this rule applies to given operation
- `production` converting given operation to iterable of one or more operations. Composing this sequence of operations should give an operation equivalent to the original one.

The engine provides generic `decompose_operations` function that can be used for constructing decomposer for (in principle) arbitrary third-party circuit library. Also, this PR provides decomposer for cirq circuits (`decompose_cirq_circuit`) and a single`PowerGateToPhaseAndRotation` decomposition rule. Depending on the configuration, this rule decomposes `YPowGate` and `XPowGate` instances that are not rotations or Pauli gates into a sequence:
```
PHASE(phi) * X * PHASE(phi) * X * ROTATION(theta)
```
where `theta` is the same as rotation angle in original gate, and `phi`, controlling global phase, is choosen such that the decomposition results in the same unitary as the original operation (see tests).


src/python/zquantum/core/circuits/_decomposition.py
```diff
@@ -0,0 +1,42 @@
+from typing import Iterable, TypeVar
+
+from typing_extensions import Protocol
+
+OperationType = TypeVar("OperationType")
+
+
+class DecompositionRule(Protocol[OperationType]):
+    def production(self, operation: OperationType) -> Iterable[OperationType]:
+        raise NotImplementedError()
```
alexjuda, line 10: WDYT about using "the `@abstractmethod` trick" so that codecov doesn't complain?

src/python/zquantum/core/circuits/_decomposition.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def predicate(self, operation: OperationType) -> bool:
+        raise NotImplementedError()
+
+
+def decompose_operation(
+    operation: Iterable[OperationType],
+    decomposition_rules: Iterable[DecompositionRule[OperationType]],
+):
+    rules_iter = iter(decomposition_rules)
```
alexjuda, line 20: I remember that forking iterators with [itertools.tee](https://docs.python.org/3/library/itertools.html#itertools.tee) is prone to performance problems if the forked iterators diverge much. I'm worried we might face a similar issue here when decomposing big circuits. WDYT?

src/python/zquantum/core/circuits/__init__.py
```diff
@@ -141,7 +141,9 @@
     builtin_gate_by_name,
 )
 from ._circuit import Circuit
+from ._cirq_decompositions import PowerGateToPhaseAndRotation, decompose_cirq_circuit
 from ._compatibility import new_circuit_from_old_circuit
+from ._decomposition import DecompositionRule, decompose_operations
```
alexjuda, line 146: The brand new decomposition engine seems to be unrelated to our Z-Quantum circuits framework. WDYT about moving it to a different place in the project, e.g. `zquantum.core.decomposition`?

alexjuda (changes requested):

src/python/zquantum/core/circuits/_decomposition.py
dexter2206: Actually, it still does complain. I already tried this trick (see commit c535a50 and a18321f that reverts it). I think we would need to make the whole protocol an abstract class. Or, we could use `pass` instead of rising `NotImplementedError`. WDYT?

src/python/zquantum/core/circuits/_decomposition.py
dexter2206: Switch to using eager slices in 374c308

src/python/zquantum/core/circuits/_decomposition.py
alexjuda: The "trick" consists of marking methods with `@abstractmethod` decorator (it raises errors on its own) and removing their implementations, either replacing with `pass` or `...`. Would you try that?

dexter2206: Hmm, I thought that the second part is unnecessary, but I'll try doing that!

src/python/zquantum/core/circuits/__init__.py
dexter2206: Moved to `zquantum.core.decomposition` in 36913a9

src/python/zquantum/core/decompositions/_decomposition.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def decompose_operation(
+    operation: Iterable[OperationType],
+    decomposition_rules: Sequence[DecompositionRule[OperationType]],
+):
+    if not decomposition_rules:
+        return [operation]
+
+    current_rule, remaining_rules = decomposition_rules[0], decomposition_rules[1:]
```
alexjuda, line 31: FYI, there's a destructuring syntax for this exact purpose :)
```suggestion
    current_rule, *remaining_rules = decomposition_rules
```

alexjuda (approved):

src/python/zquantum/core/decompositions/_decomposition.py
dexter2206: I thought it is less known, but on the other hand it's much cleaner. Done in  e69e23e

PR #379, opened 2021-06-24
Change estimation tests to use SymbolicBackend.

mstechly:


dexter2206 (approved):

src/python/zquantum/core/interfaces/mock_objects.py
```diff
@@ -26,77 +27,12 @@ def __init__(self, n_samples: Optional[int] = None):
...
-        super(MockQuantumBackend, self).run_circuit_and_measure(circuit)
-        measurements = Measurements()
-
-        n_samples_to_measure: int
-        if isinstance(n_samples, int):
-            n_samples_to_measure = n_samples
-        elif isinstance(self.n_samples, int):
-            n_samples_to_measure = self.n_samples
+        super(MockQuantumBackend, self).run_circuit_and_measure(circuit, n_samples)
+        if n_samples is not None:
```
alexjuda, line 31: Is there a reason why we have this `if` here instead of using the logic inside the delegate? 

In other words, why not just create the `SymbolicSimulator` in this class' `__init__` and proxy all method invocations to it?

alexjuda (changes requested):

src/python/zquantum/core/interfaces/mock_objects.py
mstechly: This is because we have very bad design, where `n_samples` can be coming both as argument to `run_circuit_and_measure` and as class argument. And what you suggested what exactly what I did at first, but it didn't work, as some test change `n_samples` after the object is created.
This is a known issue which we plan to fix, by getting rid of the `n_samples` in init and just leaving the one in `run_circuit_and_measure`. 
However, for the time being I've added this logic, as I couldn't figure out a better way to make it work.

alexjuda (approved):

PR #381, opened 2021-06-30
Kwargs can now be passed to estimator pre-processing

jgonthier:
Not quite sure if or how I should test that.

### DIFF
--- a/steps/measure.py
+++ b/steps/measure.py
@@ -121,8 +121,11 @@ def evaluate_ansatz_based_cost_function(
     noise_model: Optional[str] = None,
     device_connectivity: Optional[str] = None,
     prior_expectation_values: Optional[str] = None,
-    estimation_tasks_transformations_kwargs: Optional[Dict] = {},
+    estimation_tasks_transformations_kwargs: Optional[Dict] = None,
 ):
+    # Empty dict as default is bad
+    if estimation_tasks_transformations_kwargs is None:
+        estimation_tasks_transformations_kwargs = {}
     ansatz_parameters = load_array(ansatz_parameters)
     # Load qubit op
     if isinstance(target_operator, str):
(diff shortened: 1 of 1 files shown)


steps/measure.py
```diff
@@ -120,6 +120,7 @@ def evaluate_ansatz_based_cost_function(
     noise_model: Optional[str] = None,
     device_connectivity: Optional[str] = None,
     prior_expectation_values: Optional[str] = None,
+    **kwargs
```
mstechly, line 123: Since `kwargs` don't necessarily belong to `estimation_tasks_transformations`, I'd rather have a separate field `estimation_tasks_transformations_kwargs`, WDYT?

jgonthier: Fixed in 8d523d33d3d7eae1bc7fe71b7e816aa1624b05e1, at least I hope the syntax is right...

steps/measure.py
```diff
@@ -121,6 +121,7 @@ def evaluate_ansatz_based_cost_function(
     noise_model: Optional[str] = None,
     device_connectivity: Optional[str] = None,
     prior_expectation_values: Optional[str] = None,
+    estimation_tasks_transformations_kwargs: Optional[Dict] = {},
```
mstechly, line 124: You shouldn't use an empty dict (or list) as a default (see SO ). Long story short – it leads to errors that might be hard to track.
Its better to do:
```
def fun(params=None):
    if params is None:
        params = {}
```

mstechly: Should have answered in that conversation, sorry!

jgonthier: Ooh thanks for the tip! Fixed in 31f2a9ade7f1581587315afe197e356598114812

mstechly (approved):

PR #385, opened 2021-07-05
Implement cost function preprocessing

dexter2206:


src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -56,6 +57,11 @@ def _minimize(
         """
         raise NotImplementedError
 
+    def _preprocess_cost_function(
+        self, cost_function: Union[CallableWithGradient, Callable]
```
mstechly, line 61: Please add a docstring.

mstechly (changes requested):

mstechly (approved):

PR #391, opened 2021-07-15
Fix Qiskit conversions

dexter2206:
In a recent release, Qiskit switched to using symengine instead of sympy. This PR updates our code for Qiskit imports to follow this change. As a side effect, this PR also adds functions for converting symengine expressions to intermediate expression trees and a dialect for converting expression tree to symengine expression.

### DIFF
--- a/src/python/zquantum/core/circuits/symbolic/qiskit_expressions.py
+++ b/src/python/zquantum/core/circuits/symbolic/qiskit_expressions.py
@@ -13,7 +13,7 @@ from numbers import Number
 import qiskit
 
 from .expressions import ExpressionDialect, reduction
-from .sympy_expressions import expression_from_sympy
+from .symengine_expressions import expression_from_symengine
 
 
 @singledispatch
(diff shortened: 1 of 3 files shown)


src/python/zquantum/core/circuits/symbolic/qiskit_expressions.py
```diff
@@ -36,8 +36,8 @@ def _expr_from_qiskit_param_expr(
     # At the moment of writing this the qiskit version that we use (0.23.2) as well
     # as the newest version 0.23.5) does not provide a better way to access symbolic
     # expression wrapped by ParameterExpression.
-    sympy_expr = qiskit_expr._symbol_expr
-    return expression_from_sympy(sympy_expr)
+    symengine_expr = qiskit_expr._symbol_expr
+    return expression_from_symengine(symengine_expr)
```
alexjuda, line 40: Does changing this mean that from now on we'll be compatible with only the newest Qiskit version (`0.28`?) and not with the previous ones? If that's the case, WDYT about adding backwards compatibility with older Qiskit versions that use `sympy` instead of `symengine`?

alexjuda (changes requested):

src/python/zquantum/core/circuits/symbolic/qiskit_expressions.py
dexter2206: Good catch. At some point, it would be great to merge sympy/symengine translations into one coherent set of functions, which would make `core` compatible with current and older versions of qiskit. However, for now, I suggest a simple route in which we just try both of the translations and see if one worked. It is implemented in 7174e33, WDYT?

alexjuda: Makes sense!

alexjuda (approved):

PR #392, opened 2021-07-23
Updated qiskit version and added symengine dependency.

mstechly:


setup.py
```diff
@@ -59,8 +59,9 @@ def _read_version():
         "lea>=3.2.0",
         "pyquil~=2.25",
         "cirq>=0.9.1,<=0.10",
-        "qiskit~=0.26",
+        "qiskit~=0.28",
         "overrides~=3.1",
+        "symengine~=0.7",
```
alexjuda, line 64: does z-quantum-core depend directly on this library? I couldn't find any references to it in our code

alexjuda (changes requested):

setup.py
mstechly: On master – no.
On dev: https://github.com/zapatacomputing/z-quantum-core/blob/dev/src/python/zquantum/core/circuits/symbolic/symengine_expressions.py

alexjuda: Right!

alexjuda (approved):

PR #396, opened 2021-08-03
Created test cases for U3 gate in PyQuil 

Shiro-Raven:


tests/zquantum/core/circuits/conversions/pyquil_conversions_test.py
```diff
@@ -9,10 +9,33 @@
...
+SYMPY_PHI = sympy.Symbol("phi")
+SYMPY_LAMBDA = sympy.Symbol("lambda")
 SYMPY_GAMMA = sympy.Symbol("gamma")
+
 QUIL_THETA = pyquil.quil.Parameter("theta")
 QUIL_GAMMA = pyquil.quil.Parameter("gamma")
 
+"""
+Note: Those differently named Symbols/Parameters are needed due
+to the manner the conversion logic abstracts the names of the Symbols
```
alexjuda, line 21: I'm not sure I get it, could you elaborate? :)

src/python/zquantum/core/circuits/_matrices.py
```diff
@@ -117,13 +117,10 @@ def phase_matrix(angle):
...
-    https://github.com/quantumlib/Cirq/blob/292080453e22e91dc5658a0cfa5043539944a950/cirq/circuits/qasm_output.py#L70
+    https://qiskit.org/documentation/stubs/qiskit.circuit.library.U3Gate.html
+    Please note that this formulation introduces a global phase
     """
-    return (
-        rz_matrix(phi % (2 * sympy.pi))
-        * ry_matrix(theta % (2 * sympy.pi))
-        * rz_matrix(lambda_ % (2 * sympy.pi))
-    )
+    return rz_matrix(phi) * ry_matrix(theta) * rz_matrix(lambda_)
```
alexjuda, line 123: The Qiskit formula features 5 gates, whereas ours uses 3. Are these two formulas still matching?

Shiro-Raven: Qiskit sandwiches an RZ between two RX gates with constant pi/2 rotations. This is equivalent to having a single RY gate with the rotation of the RZ. It's a nice property of rotational gates.

Shiro-Raven: That's also how Cirq implements it internally [here](https://github.com/quantumlib/Cirq/blob/292080453e22e91dc5658a0cfa5043539944a950/cirq/circuits/qasm_output.py#L70)

tests/zquantum/core/circuits/conversions/pyquil_conversions_test.py
Shiro-Raven: when exporting circuits with more than 1 parameter from z-quantum to pyquil, it changes the symbol names from (alpha, beta, gamma, etc...) to (theta_0, theta_1, theta_2, etc...), so to account for that, the name of PyQuil Parameters need to be adapted accordingly to pass the tests, since we are comparing code using assembly code.

alexjuda (approved):

PR #397, opened 2021-08-03
Add seed to sample from wavefunction

laurgao:


alexjuda (commented):
@laurgao The changes Look Good To Me, but why is the base branch `feature/ZQS-359/shiro-raven/estimators-tests-as-contracts` and not `dev`? :)

laurgao:
@alexjuda ah that was because the commits from `feature/ZQS-359/shiro-raven/estimators-tests-as-contracts` are on this branch, + the commit here changed a file created in the other branch we made. so we thought we'd merge this branch into the other branch and then merge `feature/ZQS-359/shiro-raven/estimators-tests-as-contracts` into branch. might've not been the best way to do it 😅

PR #403, opened 2021-08-10
Unifying Global Phase of U3 gate and creating U3 zquantum decompositions, and Fixing bug with PyQuil Conversions

Shiro-Raven:


src/python/zquantum/core/decompositions/_zquantum_decompositions.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def production(self, operation: GateOperation) -> Iterable[GateOperation]:
+        theta, phi, lambda_ = operation.params
+
+        decomposition = [RZ(phi), RY(theta), RZ(lambda_)]
+
+        preprocessing = (
+            lambda x: x.controlled(operation.gate.num_control_qubits)
+            if operation.gate.name == "Control"
+            else x
```
dexter2206, line 34: In most of our codebase, we don't use lambdas. Instead, we typically use locally defined functions. WDYT about doing it here?

tests/zquantum/core/circuits/conversions/qiskit_conversions_test.py
```diff
@@ -108,7 +96,46 @@ def test_matrices_are_equal_up_to_phase_factor(self, theta, phi, lambda_):
...
+def _reorder_gate(G, perm):
+    perm = list(perm)
+
+    # number of qubits
+    n = len(perm)
+
+    # reorder both input and output dimensions
+    perm2 = perm + [n + i for i in perm]
+
+    return np.reshape(np.transpose(np.reshape(G, 2 * n * [2]), perm2), (2 ** n, 2 ** n))
```
dexter2206, line 111: This function is rather hard to digest and takes a while to understand. What do you think about writing some docstring explaining what it does? Also, I recommend changing its name to `_reorder_gate_matrix` since it operates on matrices, not gates.

dexter2206 (approved):

PR #404, opened 2021-08-10
ZQS 617 parameter preprocessors

dexter2206:
This PR is the first step in the cost function refactor. The implemented changes include:
- cost-function related protocols
- two implementations of `ParameterPreprocessor` protocol: `add_noise` and `fix_parameters`

### DIFF
--- a/src/python/zquantum/core/cost_function.py
+++ b/src/python/zquantum/core/cost_function.py
@@ -284,7 +284,7 @@ class CostFunction(Protocol):
 
     @abc.abstractmethod
     def __call__(self, params: np.ndarray) -> Union[float, ValueEstimate]:
-        raise NotImplementedError()
+        """Compute  value of the cost function for given parameters."""
 
 
 class EstimationTasksFactory(Protocol):
@@ -296,7 +296,7 @@ class EstimationTasksFactory(Protocol):
 
     @abc.abstractmethod
     def __call__(self, parameters: np.ndarray) -> List[EstimationTask]:
-        raise NotImplementedError()
+        """Produce estimation tasks for given parameters."""
 
 
 class ParameterPreprocessor(Protocol):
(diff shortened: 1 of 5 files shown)


src/python/zquantum/core/cost_function.py
```diff
@@ -275,3 +277,72 @@ def __call__(self, parameters: np.ndarray) -> ValueEstimate:
         )
 
         return sum_expectation_values(combined_expectation_values)
+
+
+class CostFunction(Protocol):
```
mstechly, line 282: I think these protocols should actually go to the `interfaces` dir.

src/python/zquantum/core/cost_function.py
```diff
@@ -275,3 +277,72 @@ def __call__(self, parameters: np.ndarray) -> ValueEstimate:
...
+    modifying passed parameters in place, which can have unpredictable
+    side effects.
+    """
+
+    @abc.abstractmethod
+    def __call__(self, parameters: np.ndarray) -> np.ndarray:
+        """Preprocess parameters."""
+
+
+def fix_parameters(fixed_parameters: np.ndarray) -> ParameterPreprocessor:
```
mstechly, line 315: Shouldn't it actually prepend instead of append?
For LBL training for example we:
- train a layer –> you have some good parameters for layer 1
- add a new layer
- use new parameter vector – good params from layer 1 + some new random params.

src/python/zquantum/core/cost_function.py
```diff
@@ -275,3 +277,72 @@ def __call__(self, parameters: np.ndarray) -> ValueEstimate:
...
+        preprocessor
+    """
+
+    def _preprocess(parameters: np.ndarray) -> np.ndarray:
+        return combine_ansatz_params(parameters, fixed_parameters)
+
+    return _preprocess
+
+
+def add_noise(parameter_precision, parameter_precision_seed) -> ParameterPreprocessor:
```
mstechly, line 330: Perhaps rename `add_normal_noise` to be more specific?

mstechly (approved):

PR #406, opened 2021-08-10
ZQS 671 Cost function refactor

laurgao:
This PR is the 2nd step in the cost function refactor. The implemented changes include:
- Cost function factory
- One implementation of an `EstimationTasksFactory` generator: `substitution_based_estimation_tasks_factory`

PR #407, opened 2021-08-10
ZQS 671 Dynamic circuit estimation task factory

laurgao:
3rd PR in cost function refactor

Merge after #406


src/python/zquantum/core/cost_function.py
```diff
@@ -429,3 +429,54 @@ def _tasks_factory(parameters: np.ndarray) -> List[EstimationTask]:
...
+        An EstimationTasksFactory object.
+    """
+
+    def _tasks_factory(parameters: np.ndarray) -> List[EstimationTask]:
+
+        # NOTE: `ansatz._generate_circuit(parameters)` currently does not produce an
+        # executable circuit, but rather a parametrized circuit with sympy symbols in
+        # some ansatz implementations. (Ex. see ansatzes in z-quantum-qaoa)
+        #
+        # Combined with how this is a private method, we will probably have to somewhat
```
mstechly, line 464: I would write TODO here and perhaps add a story to Jira :)

src/python/zquantum/core/interfaces/ansatz.py
```diff
@@ -87,6 +87,9 @@ def _generate_circuit(self, params: Optional[np.ndarray] = None) -> Circuit:
         Will return parametrized circuits if no parameters are passed and the ansatz
         supports parametrized circuits.
 
+        Ansatzes that implement this method must not ignore parameters if parameters
```
mstechly, line 90: I think this is a pretty logical thing, so I'm not really sure if it's worth mentioning here.

mstechly (approved):

src/python/zquantum/core/interfaces/ansatz.py
laurgao: Some ansatzes don't do it though 🤔

mstechly: Then leave it be 🤷

PR #409, opened 2021-08-23
draft for 'human' sorting which fixes symbol name miss match

MSRudolph:
The sorting of Sympy symbols did not work and sorted `theta_0, theta_1, theta_2, ... , theta_10 ` like `theta_0, theta_1, theta_10, theta_2, ...`.
This is not pretty but it fixes the problem.

The previous miss-match caused `qe-tequila` to calculate the gradients incorrectly because it relied on correct sorting of the symbol names.

### DIFF
--- a/src/python/zquantum/core/cost_function.py
+++ b/src/python/zquantum/core/cost_function.py
@@ -31,11 +31,26 @@ from .measurement import (
 )
 from .utils import ValueEstimate, create_symbols_map
 
+import re
+
+
+def check_for_digit_in_string(text):
+    return int(text) if text.isdigit() else text
+
+
+def natural_keys(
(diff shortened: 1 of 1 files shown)


src/python/zquantum/core/cost_function.py
```diff
@@ -31,11 +31,26 @@
...
 from .utils import ValueEstimate, create_symbols_map
 
+import re
+
+
+def check_for_digit_in_string(text):
+    return int(text) if text.isdigit() else text
+
+
+def natural_keys(
```
alexjuda, line 41: Would it be possible to add tests with those examples from our conversation on Slack?

alexjuda (changes requested):

PR #411, opened 2021-08-24
Enable black style checks

laurgao:
Style check actions v1.4.0 uses the same black version as our pre commit hooks, while the older one doesn't. So, upgrading to this new version allows us to enable black style checks whereas it was previously disabled.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -40,7 +40,7 @@ jobs:
           use-pylint: false
           use-pycodestyle: false
           use-flake8: true
-          # We're not ready yet to validate mypy, black & isort on every PR. In the
+          # We're not ready yet to validate mypy on every PR. In the
           # mean time, please rely on local pre-commit hook checks.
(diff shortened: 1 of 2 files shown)


.github/workflows/style.yml
```diff
@@ -32,25 +32,25 @@ jobs:
...
           # A list of all paths to test
           python-root-list: "src/python/zquantum/core steps tests/zquantum/core tests/acceptance_tests"
 
           use-pylint: false
           use-pycodestyle: false
           use-flake8: true
-          # We're not ready yet to validate mypy, black & isort on every PR. In the 
+          # We're not ready yet to validate mypy on every PR. In the
           # mean time, please rely on local pre-commit hook checks.
           use-mypy: false
```
alexjuda, line 45: There has been a point in time when z-quantum-core was also mypy-compatible. Would it make sense to at least try enabling this check, too?

alexjuda (approved):

PR #413, opened 2021-08-27
Extracting Qiskit and Symengine from core to qe-qiskit

Shiro-Raven:


alexjuda (approved):
I guess this would have to wait with merging until https://github.com/zapatacomputing/qe-qiskit/pull/63 is ready, right?

PR #416, opened 2021-08-31
ZQS 772 Update ground state cost function

laurgao:


src/python/zquantum/core/cost_function.py
```diff
@@ -94,6 +94,11 @@ def get_ground_state_cost_function(
     Returns:
         Callable
     """
+    DeprecationWarning(
```
mstechly, line 97: You need to add `warnings.warn(DeprecationWarning(...))` in order to make it work. Same in other places.

src/python/zquantum/core/cost_function.py
```diff
@@ -325,6 +333,7 @@ def create_cost_function(
     estimation_tasks_factory: EstimationTasksFactory,
     estimation_method: EstimateExpectationValues = _by_averaging,
     parameter_preprocessors: Iterable[ParameterPreprocessor] = None,
+    gradient_function: Callable = finite_differences_gradient,
```
mstechly, line 336: More specific typing would be nice.

src/python/zquantum/core/cost_function.py
```diff
@@ -338,6 +347,9 @@ def create_cost_function(
         parameter_preprocessors: a list of callable functions that are applied to
             parameters prior to estimation task evaluation. These functions have to
             adhere to the ParameterPreprocessor protocol.
+        gradient_function: a function which returns a function used to compute the
```
mstechly, line 350: Why did you add `gradient_function` to this function?

src/python/zquantum/core/cost_function.py
```diff
@@ -379,25 +391,25 @@ def _cost_function(parameters: np.ndarray) -> Union[float, ValueEstimate]:
...
     target_operator: SymbolicOperator,
-    ansatz: Ansatz,
+    parametrized_circuit: Circuit,
     estimation_preprocessors: List[EstimationPreprocessor] = None,
 ) -> EstimationTasksFactory:
     """Creates a EstimationTasksFactory object that can be used to create
-    estimation tasks dynamically with parameters provided on the fly. These
-    tasks will evaluate the parametric circuit of an ansatz, using a symbol-
-    parameter map. Wow, a factory for factories! This is so meta.
+    estimation tasks that return the estimated expectation value of the input
```
mstechly, line 403:
```suggestion
    estimation tasks that returns the estimated expectation value of the input
```

tests/zquantum/core/cost_function_test.py
```diff
@@ -33,6 +35,10 @@
 
 
 class TestGroundStateCostFunction:
+    @pytest.fixture(params=["old", "new"])
+    def factory(self, request):
```
mstechly, line 39: I would at least change it to `factory_type`.

mstechly (changes requested):

src/python/zquantum/core/cost_function.py
laurgao: To keep it consistent with other cost function factories bc other ones all take a `gradient_function` arg - do you think that's a good idea?

mstechly: Ok, makes sense!

mstechly (approved):
Looks good to me!

PR #418, opened 2021-09-01
ZQS-675 NestedOptimizer interface

laurgao:
MetaOptimizers are optimizers that modify cost function throughout optimization – an example of that would be RQAOA or LayerwiseOptimizer.


tests/zquantum/core/interfaces/optimizer_test.py
```diff
@@ -19,3 +24,37 @@ def test_cost_function_is_preprocessed_before_minimization():
     result = optimizer.minimize(simple_cost_function, np.zeros(4))
 
     assert result.opt_value == 2
+
+
+class MockMetaOptimizer(MetaOptimizer):
```
mstechly, line 29: Move this to `mock_objects.py`

tests/zquantum/core/interfaces/optimizer_test.py
```diff
@@ -19,3 +24,37 @@ def test_cost_function_is_preprocessed_before_minimization():
...
+    return EstimationTask(
+        IsingOperator("Z0 Z1", 5), ansatz.parametrized_circuit, number_of_shots=None
+    )
+
+
+def simple_cost_function_factory(estimation_tasks_factory):
+    return simple_cost_function
+
+
+def test_meta_optimizer_initializes_shit_properly_to_bring_up_code_coverage():
```
mstechly, line 49: We can remove this.

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,50 @@ def construct_history_info(cost_function, keep_history):
...
+        self,
+        ansatz: Ansatz,
+        inner_optimizer: Optimizer,
+        estimation_tasks_factory: Callable[[Any], EstimationTasksFactory],
+        cost_function_factory: Callable[[EstimationTasksFactory], CostFunction],
+        recorder: RecorderFactory = _recorder,
+    ) -> None:
+        """
+        Optimizers that modify cost function throughout optimization.
+        See RQAOA (in zquantum.qaoa) or LayerwiseAnsatzOptimizer (in
```
mstechly, line 122: We should not refer to packages that we depend on, because if something changes in `zquantum.qaoa`, then  `core` wouldn't know that.
I'd suggest creating a slightly more involved `MockMetaOptimizer` that could be referenced here and could be treated as an example.

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,50 @@ def construct_history_info(cost_function, keep_history):
     if keep_history and hasattr(cost_function, "gradient"):
         histories["gradient_history"] = cost_function.gradient.history
     return histories
+
+
+class MetaOptimizer(ABC):
+    def __init__(
+        self,
+        ansatz: Ansatz,
```
mstechly, line 114: This probably shouldn't be a part of the interface.
Ansatz could incorporated in `estimation_tasks_factory` / `cost_function_factory` or in `**kwargs`.

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,50 @@ def construct_history_info(cost_function, keep_history):
...
+        inner_optimizer: Optimizer,
+        estimation_tasks_factory: Callable[[Any], EstimationTasksFactory],
```
mstechly, line 116: Same as Ansatz.

src/python/zquantum/core/interfaces/optimizer.py
mstechly: Probably **kwargs.

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,50 @@ def construct_history_info(cost_function, keep_history):
...
+        cost_function_factory: Callable[[EstimationTasksFactory], CostFunction],
```
mstechly, line 117: I think that `cost_function_factory` should be able to take any inputs and it should be specified by particular MetaOptimizer implementation. Therefore the following signature might be better:
```suggestion
        cost_function_factory: Callable[[...], CostFunction],
```

However, I'm not sure if that's actually syntactically correct, @alexjuda could you take a look?

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,50 @@ def construct_history_info(cost_function, keep_history):
     if keep_history and hasattr(cost_function, "gradient"):
         histories["gradient_history"] = cost_function.gradient.history
     return histories
+
+
+class MetaOptimizer(ABC):
```
mstechly, line 111: Maybe a `Protocol` would be a better fit here?
I'm not sure. CC: @alexjuda

mstechly (changes requested):

src/python/zquantum/core/interfaces/optimizer.py
mstechly: Or just remove `ansatz` alltogether and leave it up to the implementation of the interface. Similarly to what we do with `Optimizer` interface, where we specify only `recorder` in the base class and then we overwrite `init` in the implementations with other inputs as well.
@alexjuda

mstechly: Talked with @dexter2206 , seems that your changes @laurgao are good, we don't want to do use `kwargs`.

src/python/zquantum/core/interfaces/optimizer.py
```diff
@@ -105,3 +106,41 @@ def construct_history_info(cost_function, keep_history):
...
+    def __init__(
+        self,
+        inner_optimizer: Optimizer,
+        cost_function_factory: Callable[..., CostFunction],
```
mstechly, line 115: `cost_function_factory` should the input to `minimize` method.

src/python/zquantum/core/interfaces/optimizer.py
alexjuda: @mstechly could you share what's the motivation of specifying `__init__` in the interface? It has a bunch of downsides, like:
- Different implementations might need different parameters in the constructor. By specyfing `__init__` in the interface we forbid any other sets of constructor parameters than the ones here.
- in this init we create object fields. If an object's class does inherit this class, the object would have these fields as well. If the object's class doesn't inherit this one, the object would have a different set of fields. This seems like a too much effects when we just want to specify common interface for the `minimize()` method.

mstechly (approved):

PR #419, opened 2021-09-01
Removal of Pyquil's conversion logic

Shiro-Raven:


tests/zquantum/core/circuits/symbolic/translations_test.py
```diff
@@ -14,126 +9,33 @@
...
-        (
-            -5 * sympy.Symbol("x") * sympy.Symbol("y"),
-            -5 * quil.Parameter("x") * quil.Parameter("y"),
-        ),
+        sympy.Symbol("theta"),
+        sympy.Symbol("theta") * sympy.Symbol("gamma"),
+        sympy.cos(sympy.Symbol("theta")),
+        sympy.cos(2 * sympy.Symbol("theta")),
+        sympy.exp(sympy.Symbol("x") - sympy.Symbol("y")),
+        sympy.Add(
```
dexter2206, line 19: Previously we used `sympy.Add` with `evaluate=False` to preserve the ordering of the arguments. This was needed so that comparison to pyquil expression would succeed. Maybe this is no longer needed, and we could just use + operator?

dexter2206 (changes requested):
Looks ok, but we could take a look if it is possible to get rid of all `sympy.Add` and `sympy.Mul` with `evaluate=False` in tests.

dexter2206 (approved):

PR #421, opened 2021-09-08
Remove PyQuil's Wavefunction from Core

Shiro-Raven:


src/python/zquantum/core/wavefunction.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+
+        if isinstance(amplitude_vector, np.ndarray) or isinstance(
+            amplitude_vector, list
+        ):
+            amplitude_vector = Matrix(amplitude_vector)
+
+        self._check_sanity(amplitude_vector)
+
+        self._wavefunction = amplitude_vector
```
mstechly, line 43: Why not actually call it `_amplitude_vector` instead of`_wavefunction` ?

src/python/zquantum/core/wavefunction.py
```diff
@@ -0,0 +1,142 @@
...
+    A simple wavefunction data structure that can
+    be used to calculate amplitudes of quantum states.
+
+    Args:
+        amplitude_vector: the initial amplitudes of the system,
+            can either be a NumPy ndarray or a SymPy Matrix
+    """
+
+    def __init__(
+        self, amplitude_vector: Union[np.ndarray, Matrix, List[complex]]
```
mstechly, line 29: Why sympy Matrix and not Vector?

src/python/zquantum/core/wavefunction.py
```diff
@@ -0,0 +1,142 @@
...
+def _cast_sympy_matrix_to_numpy(sympy_matrix, complex=False):
+    new_type = np.complex128 if complex else np.float64
+
+    try:
+        return np.array(sympy_matrix, dtype=new_type).flatten()
+    except TypeError:
+        return np.array(sympy_matrix, dtype=object).flatten()
+
+
+class Wavefunction:
```
mstechly, line 18: Overall I'm afraid that insisting on using sympy wavefunctions might kill the performance.
I don't have a good solution for that, but I think it's worth having a discussion about it.
I mean – in general in the whole codebase, not only the wavefunctions themselves.
CC: @dexter2206 @alexjuda

tests/zquantum/core/wavefunction_test.py
```diff
@@ -0,0 +1,193 @@
...
+    @pytest.mark.parametrize(
+        "obj", [Wavefunction.init_system(2), Wavefunction([1.0, 0.0])]
+    )
+    def test_init_returns_sympy_matrix(self, obj: Wavefunction):
+        assert isinstance(obj._wavefunction, Matrix)
+
+    @pytest.mark.parametrize(
+        "input_list", [[], np.zeros(17), create_random_wavefunction(3)[:-1]]
+    )
+    def test_init_fails_when_passed_list_is_not_power_of_two(self, input_list):
```
mstechly, line 23:
```suggestion
    def test_init_fails_when_len_of_passed_list_is_not_power_of_two(self, input_list):
```

mstechly (commented):
I think this will be pretty crucial part of the code, so I'd like to ask @dexter2206 for an additional review.
Also – before merging it we need to check whether we need to change all the backends to conform to the new interface or do they simply inherit after the base class and don't use pyquil wavefunctions explicitly.
Happy to help with that.

src/python/zquantum/core/wavefunction.py
Shiro-Raven: Vectors in Sympy are designed for the purpose of spatial coordinates and not as a generic mathematical vector.

More can be seen [here](https://docs.sympy.org/latest/modules/vector/index.html).

mstechly: Good to know, thanks!

src/python/zquantum/core/wavefunction.py
Shiro-Raven: @mstechly I am just trying to avoid using the exact naming as PyQuil. If you think the naming is weird, I can change it

src/python/zquantum/core/wavefunction.py
alexjuda: I thought about this too, but didn't want to fall for optimization without running any benchmarks. Do you feel like we should prepare some experiments and measure the performance before going with this refactor?

Shiro-Raven: my only argument for using sympy is the ability to calculate the final output of the circuit only once and never again, since we can depend on Symbol substitutions instead of repeating matrix multiplications for every run. This will be specially useful when running deep circuits.

mstechly: @alexjuda – I don't think it should be a blocker and also I feel that it's a more general problem with other places in our codebase as well. Will create a ticket for that.

@Shiro-Raven I haven't thought about it, but that could be actually very efficient!

src/python/zquantum/core/wavefunction.py
mstechly: Please change :)

src/python/zquantum/core/wavefunction.py
dexter2206: @mstechly @alexjuda We decided with @Shiro-Raven  to use similar approach as when converting circuits to unitaries. In current implementation we use numpy objects if possible and only fallback to sympy if there are any free  symbols presents in wavefunction.

The decision was made to make sure that the interface and inner workings of our wavefunction are precisely the same  as the PyQuil's wavefunction.

src/python/zquantum/core/wavefunction.py
```diff
@@ -0,0 +1,149 @@
...
+        except ValueError:
+            self._wavefunction[idx] = old_val
+
+            raise ValueError("This assignment violates probability unity.")
+
+    def __str__(self) -> str:
+        return self._wavefunction.__str__()
+
+    def __eq__(self, other) -> bool:
+        return self._wavefunction == other._wavefunction
```
dexter2206, line 115: This implementation will fail if the wavefunction is compared to something that is not a wavefunction. It would be more pythonic to return  `False` instead of raising an exception in such cases.

Also, the `__eq__` method seems to not be covered by tests.  I imagine we should have positive and negative cases for wfs with/without free symbols and some negative cases for wfs with/without free symbols and a negative case for comparing wavefunction to a list.

dexter2206 (changes requested):
Requested some minor changes

src/python/zquantum/core/hamiltonian.py
```diff
@@ -38,7 +38,7 @@ def group_comeasureable_terms_greedy(
     qubit_operator: QubitOperator, sort_terms: bool = False
 ) -> List[QubitOperator]:
     """Group co-measurable terms in a qubit operator using a greedy algorithm. Adapted
-    from pyquil. Constant term is included as a separate group.
+    from PyQuil. Constant term is included as a separate group.
```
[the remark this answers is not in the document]
Shiro-Raven: The docs of this file are not consistent at all. So I guess I can try and do so.

src/python/zquantum/core/evolution.py
```diff
@@ -45,7 +45,15 @@ def time_evolution(
 
 
 def _adjust_gate_angle(operation: circuits.GateOperation, time):
-    """Adjust angle in gate operation to account for evolution in time."""
+    """Adjust angle in gate operation to account for evolution in time.
```
mstechly, line 48: Actually it looks that this function is no longer used and can be removed.

mstechly: We can also remove single dispatch for `time_evolution_for_term` now, as there's only one implementation :)

src/python/zquantum/core/wavefunction.py
```diff
@@ -89,30 +92,33 @@ def _calculate_probability_of_ground_entries(arr: Matrix) -> np.float64:
...
+            self._check_sanity(self._amplitude_vector)
         except ValueError:
-            self._wavefunction[idx] = old_val
+            self._amplitude_vector[idx] = old_val
 
             raise ValueError("This assignment violates probability unity.")
 
     def __str__(self) -> str:
-        return self._wavefunction.__str__()
+        return self._amplitude_vector.__str__()
```
mstechly, line 115: Codecov complains that this is not covered by tests.

src/python/zquantum/core/wavefunction.py
```diff
@@ -46,23 +46,26 @@ def __init__(
...
-            self._wavefunction = Matrix(amplitude_vector)
+            self._amplitude_vector = Matrix(amplitude_vector)
 
-        self._check_sanity(self._wavefunction)
+        self._check_sanity(self._amplitude_vector)
 
     @property
     def amplitudes(self):
-        return _cast_sympy_matrix_to_numpy(self._wavefunction, complex=True)
+        if _free_symbols(self._amplitude_vector):
```
mstechly, line 57: I would change `_free_symbols` to `_check_free_symbols` or `_has_free_symbols`.

dexter2206: This function  returns free symbols, i.e. it is not a boolean function.

dexter2206 (approved):

PR #425, opened 2021-09-21
Add assertion for number of qubits in Circuit init

Shiro-Raven:


src/python/zquantum/core/circuits/_circuit.py
```diff
@@ -39,6 +39,10 @@ def __init__(
         n_qubits: Optional[int] = None,
     ):
         self._operations = list(operations) if operations is not None else []
+
+        if n_qubits and not isinstance(n_qubits, int):
+            raise ValueError("n_qubits must be an integer.")
+
```
dexter2206, line 45: I think this if won't fire if `n_qubits=0.0`. Since `n_qubits` being 0 does not make sense anyway, WDYT about also testing if `n_qubits` is nonzero?

dexter2206 (changes requested):

src/python/zquantum/core/circuits/_circuit.py
```diff
@@ -1,6 +1,7 @@
 import operator
 from functools import reduce, singledispatch
 from typing import Any, Dict, Iterable, List, Optional, Union
+from warnings import warn
```
dexter2206, line 4: Do we use this import anywhere in the code?

dexter2206 (changes requested):
I think we have an unused import in _circuit.py, otherwise looks good!

dexter2206 (approved):

PR #428, opened 2021-09-28
ZQS-257 add support for nonnative operations in QuantumSimulator interface

dexter2206:
This adds support for nonnative operations in QuantumSimulator interface.

For some context: some circuit elements (called `Operations` in `zquantum.core`) may be unsupported by some backends and simulators. While not much can be done in the case of backends, simulators could simulate parts of circuits they support, then apply the unsupported operations directly (via their `apply`) method, then switch again to native simulations, and so forth.

As a proposed solution, each concrete  `QuantumSimulator` should:
- implement `_get_wavefunction_from_native_circuit` for simulating circuits comprising only natively-supported operations. This method should basically do whatever `get_wavefunction` in original simulator did. The only difference is that `_get_wavefunction_from_native_circuit` should also accept `initial_state` from which the simulation should start.
- implement `is_supported_natively` method that serves as a predicate deciding whether a given operation is natively supported or not. By default, all `GateOperations` are considered natively supported.

The abstract `QuantumSimulator` utilizes the aforementioned methods in its `get_wavefunction` method, splits circuit into an alternating sequence of native/nonnative subcircuits, and either passes them through to the underlying simulator (native subcircuits) or sequentially applies operations directly to the state vector computed so far.

The proposed solution has the benefit of retaining the public interface of QuantumSimulators. However, the drawback is that the private machinery in the base class changes in such a way, that each simulator we have has to be adapted to it before this PR can be merged.


src/python/zquantum/core/interfaces/backend.py
```diff
@@ -120,14 +143,51 @@ def __init__(
...
+              should assume that the initial state is |0...0>, which typically does not
+              require any initialization.
+        Returns:
+            Wavefunction object encoding amplitudes of final state.
+        """
+
+    def is_natively_supported(self, operation: Operation) -> bool:
+        """Determine if given operation is natively supported by this Simulator.
+
+        This method will be used as a predicate in split_circuit function.
```
mstechly, line 165: It can be used as a predicate, but I think this is not necessarily the only use, so perhaps docstring should change?

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -120,14 +143,51 @@ def __init__(
...
+
     def get_wavefunction(self, circuit: Circuit) -> Wavefunction:
         """Returns a wavefunction representing quantum state produced by a circuit
 
         Args:
             circuit: quantum circuit to be executed.
         """
-        self.number_of_circuits_run += 1
-        self.number_of_jobs_run += 1
+        state = np.zeros(2 ** circuit.n_qubits)
```
mstechly, line 175: Perhaps it should take the initial state as an optional argument? If `None` it would evaluate to |0...0>, but otherwise it would allow people to seed the simulation with whatever initial state they want?

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -108,6 +109,28 @@ def get_bitstring_distribution(
         return measurements.get_distribution()
 
 
+def split_circuit(
```
mstechly, line 112: Actually doesn't it belong more to the `circuits` module rather than `backend` ?

src/python/zquantum/core/interfaces/backend.py
```diff
@@ -120,14 +143,51 @@ def __init__(
...
-        self.number_of_jobs_run += 1
+        state = np.zeros(2 ** circuit.n_qubits)
+        state[0] = 1
+        for is_supported, subcircuit in split_circuit(
+            circuit, self.is_natively_supported
+        ):
+            # Native subcircuits are passed through to the underlying simulator.
+            # They also count towards number of circuits and number of jobs run.
+            if is_supported:
+                self.number_of_circuits_run += 1
```
mstechly, line 183: I'm not sure whether these `+=` operations shouldn't be actually outside of this for loop?

dexter2206: I think not. If your circuit is split into 3 parts, say native -> non native -> native, you are running two different circuits using simulator.

mstechly: Fine 👌 
I would just add more info about how this is being handled somewhere to the docstrings, cause this is something that might be potentially confusing to users.

src/python/zquantum/core/interfaces/backend.py
dexter2206: Fixed in  51fd75b

src/python/zquantum/core/interfaces/backend.py
dexter2206: Great idea! Fixed in 1ce1140

src/python/zquantum/core/interfaces/backend.py
dexter2206: I tried to describe it in 081b287, let me know if you think the description is sufficient :)

src/python/zquantum/core/interfaces/backend.py
dexter2206: 3b7cc23 moves `split_circuits` into `zquantum.core.circuits`. Corresponding tests are also moved accordingly.

src/python/zquantum/core/interfaces/backend.py
mstechly: It lacks typing :)

src/python/zquantum/core/interfaces/backend.py
mstechly: Looks good 👌

mstechly (approved):

PR #429, opened 2021-09-28
Extract cirq from core

Shiro-Raven:


tests/zquantum/core/circuits/_unitary_test.py
```diff
@@ -1,49 +1,55 @@
...
+    def test_without_free_params_gives_the_same_result_as_cirq(
+        self, circuit, expected_unitary_filename
+    ):
         zquantum_unitary = circuit.to_unitary()
-        cirq_circuit = export_to_cirq(circuit)
-        cirq_unitary = cirq.unitary(cirq_circuit)
+
+        """
+        Note: We decided to go with file-based approach after extracting cirq
+        from z-quantum-core and not being able to use `export_to_cirq` anymore.
```
mstechly, line 41: And why not just use SymbolicSimulator to calculate these unitaries?

mstechly: I also don't see why we have these in text files and not in a single file where we can import it from as a variable?
+ the path shouldn't be called `cirq_unitaries`, just `unitaries`, right?

mstechly (changes requested):

tests/zquantum/core/circuits/_unitary_test.py
```diff
@@ -1,51 +1,59 @@
...
             # Cirq circuits have the same number of qubits as Zquantum ones.
-            Circuit([RX(np.pi / 5)(0)]),
-            Circuit([RY(np.pi / 2)(0), RX(np.pi / 5)(0)]),
-            Circuit([I(1), I(2), I(3), I(4), RX(np.pi / 5)(0), XX(0.1)(5, 0)]),
-            Circuit(
-                [
-                    XY(np.pi).controlled(1)(3, 1, 4),
-                    RZ(0.1 * np.pi).controlled(2)(0, 2, 1),
-                ]
+            (Circuit([RX(np.pi / 5)(0)]), 2),
```
mstechly, line 24: Can you sort these so that `unitary_index` as same as index of items in this list :) ?

mstechly (approved):

PR #432, opened 2021-10-01
Removed pip version from docker image.

mstechly:


docker/z-quantum-default/Dockerfile
```diff
@@ -33,8 +33,6 @@ RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1
 
 ENV PYTHONPATH="/usr/local/lib/python3.7/dist-packages:${PYTHONPATH}"
 
-# Downgrade pip to 20.2.4 until the issues with dependency resolving in pip is solved.
-RUN python3 -m pip install pip==20.2.4
```
alexjuda, line 37: 1. What version of pip we will end up having?
2. WDYT about upgrading pip to the newest available version at the moment of building the image?

docker/z-quantum-default/Dockerfile
```diff
@@ -63,7 +61,7 @@ RUN python3 -m pip install --no-cache git+git://github.com/zapatacomputing/z-qua
 
 # Misc libraries that we'd like to have already preinstalled.
 RUN python3 -m pip install \
-        codecarbon \
+        codecarbon==1.2.0 \
```
alexjuda, line 64: Why is this exact pin required?

docker/z-quantum-default/Dockerfile
mstechly: Ad 1. The newest one I hope?
Ad 2. You mean changing it to `RUN python3 -m pip install --upgrade pip` ?

docker/z-quantum-default/Dockerfile
mstechly: Cause the code that's using it is quite fragile and I'm afraid that changing a different version might accidentally break things.

alexjuda: Can you please add a comment to the dockerfile that describes what code exactly depends on this pin?

docker/z-quantum-default/Dockerfile
alexjuda: Ad 1. Probably not, this is why I'm asking :D I'd check myself but I'm assuming you've built the image locally at leaset once to check if your changes work, so it might be easier for you to spin a container and check the pip version. Something like `docker run --rm tag-of-your-image python3 -m pip -V`
Ad 2. Yup, this is what I had in mind

mstechly: Fixed in https://github.com/zapatacomputing/z-quantum-core/pull/432/commits/47d618e6d8737eb1936ef5b75e7f8906698670b6

docker/z-quantum-default/Dockerfile
mstechly: Fixed in https://github.com/zapatacomputing/z-quantum-core/pull/432/commits/aa545118c3f1deb3861d9f94d6d122f2726d291a

alexjuda (approved):

PR #439, opened 2021-10-22
Change strings to tuples for measurement distributions

Shiro-Raven:
This PR is the 2nd part of our efforts of standardizing the representation of basis states. The following changes are made:

0. The `bitstring_distribution` module is now simply `distributions`, and the distance_measures folder has been deleted with its contents being added to `distributions`. This has been done to comply with Zen of Python's "Flat is better than nested" as per @alexjuda's advice.
1. `BitstringDistribution` is now `DitSequenceDistribution`. The internal data structure representing the state of the system is now a tuple of non-negative integers, instead of a binary string. This allows for support of arbitrary-level systems.
2. The constructor of  `DitSequenceDistribution` supports 3 different types of input dictionaries. It accepts dictionaries whose are (not necessarily exclusively) binary strings, non-negative integer tuples, or strings of comma-separated non-negative integers. This provides a level of backward-compatibility. 
3. a. For saving distributions in JSON files, the tuples are converted to comma-separated non-negative integer strings. Loading distributions shouldn't be an issue since the constructor accepts this comma-separated input as well.
b. The dictionary key for the JSON is now `ditsequence-distribution`, and the schema string has also been similarly modified.
4. Tests and implementations of other modules have been accordingly adapted to this new implementation. Most notably, the `serialization` module has been modified to preprocess the `DitSequenceDistribution`, similar to how it's done in point **3a**.
5.  The tests of `ditsequence_distribution` have now been refactored by including related tests to test classes. In addition, the tests of input validation have been refactored to parameterized tests, taking as parameters a function and an input. One of the reasons of doing this was `flake` complaining a lot about the line length of the test function name. 
6. A wrapper module `bitstring_distribution` has been created to avoid breaking any of the scientists' code for the time being. It simply imports the stuff from `distributions`.


mstechly (changes requested):
Another thought – maybe for the ease of use and backward compatibility, we could introduce an alias `BistringDistribution`, that will be basically the same thing as `DitSequenceDistribution`, but less confusing for majority of people who are using BitStrings anyway?

mstechly (approved):

PR #446, opened 2021-11-29
Zqs 870/track backend usage

AthenaCaesura:
Created MeasurementTrackingBackend class which allows one to keep track of all the measurement data collected in an experiment in an external json file. Namely, we keep track of num_shots, bitstrings, counts for each measurement outcome, the implemented circuit, and the number of gates.

This generalized code written in qe-qiskit which only tracked backends for qiskit. This code expands the scope to any type of backend.

The measurement tracking backend class is triggered when the user puts the flag "track_backend_usage" : True in the workflow. The class wraps around an inner_backend and writes all the results of measurements to "raw_measurement_data.json" after each experiment. 

Testing for measurement tracking backend to be added in an upcoming story.


mstechly (changes requested):

.coveragerc
```diff
@@ -1,15 +0,0 @@
-# .coveragerc to control coverage.py
```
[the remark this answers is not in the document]
AthenaCaesura: I found another copy of .coveragerc in the main branch and put an identical file in my branch. Hope that's ok.

src/python/zquantum/core/trackers.py
```diff
@@ -0,0 +1,146 @@
...
+            circuit, n_samples
+        )
+
+    def record_raw_measurement_data(
+        self, circuit: Circuit, measurement: Measurements
+    ) -> None:
+        """Append data from a measurement to self.raw_measurement_data.
+
+        Args:
+            circuit (Circuit): Implemented circuit.
```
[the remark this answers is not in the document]
AthenaCaesura: Done. :)

src/python/zquantum/core/trackers.py
```diff
@@ -0,0 +1,146 @@
...
+from .interfaces.backend import QuantumBackend, QuantumSimulator
+from .measurement import Measurements
+from .utils import SCHEMA_VERSION
+
+
+class MeasurementTrackingBackend(QuantumBackend):
+    """A wrapper class for a backend that tracks all measurements. The measurements
+    are stored in the raw_circuit_data variable as a list of measurement objects.
+
+    To enable tracking, simply put the key-value pair "track_measurements" : True
```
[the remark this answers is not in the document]
AthenaCaesura: I put the comment in the docstring of all of the methods in steps where MeasurementTrackingBackend could be called. If methods have no docstrings, I just put the comment in the module docstring (i.e. `measurements.py`).

src/python/zquantum/core/trackers.py
```diff
@@ -0,0 +1,139 @@
...
+        with open(self.file_name, "w+") as f:
+            data = {
+                "schema": SCHEMA_VERSION + "-raw-measurement-data",
+                "raw-measurement-data": self.raw_measurement_data,
+            }
+            f.write(json.dumps(data))
+        self.raw_measurement_data = []
+
+    def rename_raw_measurement_data_file(self) -> None:
+        rename(self.file_name, "raw-measurement-data.json")
```
alexjuda, line 139: Do I understand correctly that the motivation for this method was to let the user to store the data under a different file name? WDYT about removing this method and allowing the user to pass a file object to `save_raw_measurement_data`?

steps/evaluation.py
```diff
@@ -68,6 +78,9 @@ def get_expectation_values_for_qubit_operator(
 
     save_expectation_values(expectation_values[0], "expectation-values.json")
 
+    if backend_specs.get("track_measurements"):
+        backend.rename_raw_measurement_data_file()
+
```
alexjuda, line 83: Shouldn't this follow invocation of `backend.save_raw_measurement_data()`? If so, can you help me find it?

Note that this comment will be out of date if you decide to merge `save...` and `rename...` methods.

tests/zquantum/core/trackers_test.py
```diff
@@ -0,0 +1,146 @@
...
+        optimizer = ScipyOptimizer(method="L-BFGS-B", options={"maxiter": 10})
+        cost_function = AnsatzBasedCostFunction(
+            H, ansatz, backend, estimation_method, estimation_preprocessors
+        )
+        initial_params = np.array([0, 0])
+        # When
+        opt_results = optimizer.minimize(cost_function, initial_params)
+        circuit = ansatz.get_executable_circuit(opt_results.opt_params)
+        backend.n_samples = 10000
+        backend.run_circuit_and_measure(circuit, n_samples=1000)
```
alexjuda, line 52: If we can pass `n_samples` to `run_circuit_and_measure`, what's the reason to mutate the backend on line above it?

tests/zquantum/core/trackers_test.py
```diff
@@ -0,0 +1,146 @@
...
+        f = open(backend.file_name)
```
alexjuda, line 53: 1. If you don't wrap `open(...)` in a context manager you're responsible for closing the file pointer on your own. I can't see the `f.close()` in this test, so I'd expect FS errors due to removal of an dangling/opened file at the test cleanup. Is there a reason not to use `with open(...) as f`?
2. If this test fails for some reason we end up with temporary files cluttering the current PWD. WDYT about using a [tempfile]()https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) or a [tempdir](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)? You can see how we use that in [SDK's tests](https://github.com/zapatacomputing/orquestra-sdk/blob/a9b2f1d94d0f7671c7bbb357acfde1ffb8dbb9ba/tests/v2/test_dispatch.py#L10-L46)

alexjuda (commented):
I've added a few questions/suggestions :) 

Side question: we still have ["acceptance tests"](https://github.com/zapatacomputing/z-quantum-core/tree/main/tests/acceptance_tests) that only contain tests for out steps. Wanna add tests for backend tracker there, too?

mstechly:
@alexjuda regarding acceptance tests – I don't think it's necessary. Since we'll be deleting them in near future, I don't think it's worth spending time on them.

mstechly (approved):
Looks good to me!
Just please add `.coveragerc` again ;)
Waiting for approval from @alexjuda

PR #448, opened 2021-12-06
Zqs 876/cov and corr tests

AthenaCaesura:
Added tests for correlation and covariance for the estimation_expectation_values_by_averaging method.

Specifically, I added tests for when we have a low number of samples and high number of samples. This is necessary as when the number of samples is high, the covariance teds to zero.

Right now it might be hard to see exactly which of the tests fail because they are all performed together. This keeps convention with other tests in the class, however I think this could be implemented better.


tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -685,11 +760,87 @@ def test_calculate_exact_expectation_values(
...
+                np.testing.assert_array_almost_equal(corr, target_corr, decimal=1)
+                # All covariances should be close to 0 when number_of_shots is high.
+                np.testing.assert_array_almost_equal(target_cov, cov, decimal=2)
+
+    @pytest.mark.parametrize(
+        "mock_estimation_tasks, target_expectations",
+        TEST_CASES_NONEIGENSTATES_WITH_LOW_NUMBER_OF_SHOTS,
+    )
+    def test_covariance_when_averaging_for_non_eigenstates_and_number_of_shots_is_low(
+        self, target_expectations, mock_estimation_tasks
```
alexjuda, line 823: Keeping the order of arguments matching the fixtures would help me read it better
```suggestion
        self, mock_estimation_tasks, target_expectations
```

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -603,11 +629,62 @@ def test_evaluate_non_measured_estimation_tasks_fails_with_non_zero_shots(
...
+TEST_CASES_NONEIGENSTATES_WITH_LOW_NUMBER_OF_SHOTS = [
+    (
+        4
+        * [
+            EstimationTask(
+                IsingOperator("[Z0] + [Z1]"),
+                circuit=Circuit([X(0), X(1)]),
+                number_of_shots=10,
+            ),
+        ],
```
alexjuda, line 652: There are 4 test examples. They seem to have the same "inputs", but different ref "outputs".

Is it because the actual input values don't matter? If so, WDYT about not hardcoding the disposable estimation tasks and passing `None` in the test instead?

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -603,11 +629,62 @@ def test_evaluate_non_measured_estimation_tasks_fails_with_non_zero_shots(
...
+
+
+# needs it's own class otherwise issues arise with calling run_circuitset_and_measure.
+class MockBackendForTestingCovariancewhenNumberOfShotsIsLow:
+    def run_circuitset_and_measure(self, circuit, shots_per_circuit):
+        return [
+            Measurements([(0, 1), (0, 0)]),
+            Measurements([(1, 0), (0, 0)]),
+            Measurements([(1, 0), (0, 1)]),
+            Measurements([(1, 1), (0, 0)] * 10),
```
alexjuda, line 686: Are these values closely tied to the values in matrices in `TEST_CASES_NONEIGENSTATES_WITH_LOW_NUMBER_OF_SHOTS`? If so, would it make sense to treat them as the "inputs" to your unit test?

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -685,11 +760,87 @@ def test_calculate_exact_expectation_values(
...
+            expectation_values_list, target_expectations, estimation_tasks
+        ):
+            for cov, corr, target_corr, target_cov in zip(
+                expectation_values.estimator_covariances,
+                expectation_values.correlations,
+                target.correlations,
+                target.estimator_covariances,
+            ):
+                assert len(corr) == len(task.operator.terms)
+                assert len(corr) == len(task.operator.terms)
```
alexjuda, line 786: These two lines are duplicated. Did you mean to assert on `cov`?
```suggestion
                assert len(corr) == len(task.operator.terms)
                assert len(cov) == len(task.operator.terms)
```

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -685,11 +760,87 @@ def test_calculate_exact_expectation_values(
...
+            for cov, corr, target_corr, target_cov in zip(
+                expectation_values.estimator_covariances,
+                expectation_values.correlations,
+                target.correlations,
+                target.estimator_covariances,
+            ):
+
+                np.testing.assert_array_almost_equal(corr, target_corr, decimal=2)
+                # All covariances should be close to 0 when number_of_shots is high.
+                np.testing.assert_array_almost_equal(cov, target_cov, decimal=2)
```
alexjuda, line 842: The only covariance close to zero I've found is the test example 4  (the matrix with `0.05` on every element). Do other test examples also fit this assertion?

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -685,11 +760,87 @@ def test_calculate_exact_expectation_values(
...
+        )
+        for expectation_values, target in zip(
+            expectation_values_list, target_expectations
+        ):
+            for cov, corr, target_corr, target_cov in zip(
+                expectation_values.estimator_covariances,
+                expectation_values.correlations,
+                target.correlations,
+                target.estimator_covariances,
+            ):
```
alexjuda, line 838: If I understand correctly this loop iterates over matrix rows. WDYT about removing it and comparing full matrices instead of going row-by-row?

alexjuda (changes requested):

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -1,10 +1,11 @@
 from functools import partial
+from math import sqrt
```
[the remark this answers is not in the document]
AthenaCaesura: Yup! just didn't catch it.

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -584,8 +585,33 @@ def test_evaluate_non_measured_estimation_tasks_fails_with_non_zero_shots(
...
+            # tests correlation
+            EstimationTask(
+                IsingOperator("[Z0] + [Z1]"),
+                circuit=Circuit([X(0), X(1)]),
+                number_of_shots=10,
+            ),
+            # tests negative correlation
+            EstimationTask(
+                IsingOperator("[Z0] + [Z1]"),
+                circuit=Circuit([I(0), X(1)]),
```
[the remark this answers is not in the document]
AthenaCaesura: I had it in my head that `run_circuit_and_measure` throws an error if no gates are enacted on a particular qubit. i rand it again without `I(0)` and it's fine. 🤷‍♀️

tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -603,11 +629,62 @@ def test_evaluate_non_measured_estimation_tasks_fails_with_non_zero_shots(
...
+                np.array([0]),
+                [np.array([[1, 1], [1, 1]])],
+                [np.array([[0.05, 0.05], [0.05, 0.05]])],
+            ),
+        ],
+    ),
+]
+
+
+# needs it's own class otherwise issues arise with calling run_circuitset_and_measure.
```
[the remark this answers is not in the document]
AthenaCaesura: Omg and I call English my first language!

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: We need to pass some sort of `estimation_task` in order to avoid errors from `run_circuit_and_measure`. I updated the docstring in my new PR to make this more clear.

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: Of course. addressed.

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: Oops! that was left over from a copy/paste I did. thanks!

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: I want the test to fail for particular jobs to that the user can tell exactly what went wrong. I think comparing these all at once would make the error output less clear.

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: Yup!

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: I'm not sure how to make these outputs for `run_circuit_and_measure` while still putting them as inputs to the test.

alexjuda: [not a blocker] How about hardcoding those right inside the test method or under a [`pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html)?

alexjuda (approved):

PR #450, opened 2021-12-14
Update setup.py

mstechly:


setup.py
```diff
@@ -51,8 +51,7 @@ def _read_readme():
         "numpy>=1.20",
         "scipy>=1.4.1",
         "sympy>=1.5",
-        # TODO: use openfermion from PyPI when 1.3 is released
-        "openfermion @ git+https://github.com/quantumlib/OpenFermion@7532ad9e6d34f6922537f00a5cc001047f2bfdd1",  # noqa: 501
+        "openfermion~=1.3",
```
alexjuda, line 54: WDYT about relaxing the pin to sth like `~1.0`? AFAIK z-quantum-core works with most openfermion versions. We only had problems when there was a project that would only work `openfermion==1.3`, but it's unrelated to zq-core as long as our requirements are loose enough

alexjuda (changes requested):

setup.py
mstechly: Or maybe just use `<=` as we did in `qe-cirq` (https://github.com/zapatacomputing/qe-cirq/blob/26b00f1c290bac07f95b4b15d06dd7874e907b14/setup.py#L30) ?
One thing I'm afraid of is that something will break with a new version, so would rather not use `~=1.0`.

alexjuda (approved):

PR #451, opened 2021-12-17
estimate_expectation_values_by_averaging allows for all tasks not measure

AthenaCaesura:
`estimate_expectation_values_by_averaging` used to throw an error which was difficult to interpret when passed a list of `ExpectationValues` none of which could be measured. This hotfix ensures that when this cases arises, we simply return all of the non-measured expectation values.


tests/zquantum/core/estimation/estimation_test.py
```diff
@@ -696,3 +696,19 @@ def test_calculate_exact_expectation_values_fails_with_non_simulator(
...
+        target = [
+            ExpectationValues(np.array([0.0]), [np.array([[0.0]])], [np.array([0.0])])
+        ]
+
+        expectation_values_list = estimate_expectation_values_by_averaging(
+            simulator, estimation_tasks
+        )
+
+        # test that all attributes are the same
+        assert expectation_values_list[0].__dict__ == target[0].__dict__
```
mstechly, line 714: We can't just compare two `ExpectationValues` objects? Maybe we should be able to do that?

mstechly: [request] Seems like a good idea to me to also check that the length of `expectation_values_list` is 1?

mstechly (approved):

src/python/zquantum/core/estimation/_estimation.py
```diff
@@ -225,22 +225,22 @@ def split_estimation_tasks_to_measure(
     """
 
     estimation_tasks_to_measure = []
-    estimation_tasks_not_measured = []
+    estimation_tasks_not_to_measure = []
     indices_to_measure = []
     indices_to_not_measure = []
```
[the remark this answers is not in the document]
AthenaCaesura: Thanks! I missed these when I was going over the code.

tests/zquantum/core/estimation/estimation_test.py
AthenaCaesura: Added a method of comparing ExpectationValues and assertion that list has length 1.

We might want to make the comparison method a little more sophisticated than what I have written.

PR #452, opened 2021-12-19
ZQS 890 Improve check parity performance with vectorization

laurgao:


src/python/zquantum/core/measurement.py
```diff
@@ -412,48 +428,69 @@ def convert_bitstring_to_int(bitstring: Sequence[int]) -> int:
     return int("".join(str(bit) for bit in bitstring[::-1]), 2)
 
 
-def check_parity(
-    bitstring: Union[str, Sequence[int]], marked_qubits: Iterable[int]
-) -> bool:
-    """Determine if the marked qubits have even parity for the given bitstring.
+def check_parity_of_vector(
```
mstechly, line 431: [request] You shouldn't replace `check_parity` method with some other, as it breaks API – libraries depending on `z-quantum-core` might be using it. I would just add this as a new method.

mstechly: Also, it is called `check_parity_of_vector`, but actually doesn't really take in vector – (at least I wouldn't call "list of strings" a vector).
I think in the case of this method, since speed is important, it's better to just say that it takes only one type of inputs (probably `np.ndarray`) which is efficient.

src/python/zquantum/core/measurement.py
```diff
@@ -347,31 +347,47 @@ def get_parities_from_measurements(
...
-        values.append([0, 0])
         marked_qubits = [op[0] for op in term]
-        for bitstring, count in bitstring_frequencies.items():
-            if check_parity(bitstring, marked_qubits):
-                values[-1][0] += count
-            else:
-                values[-1][1] += count
+        parity = check_parity_of_vector([*bitstring_frequencies.keys()], marked_qubits)
+
+        bitstrings_with_true_parity: int = (parity * bitstring_counts).sum()
```
mstechly, line 360: [suggestion] This variable is not really `bitstrings` –  it's an integer.
I would suggest renaming it to `true_parity_count` or sth like that.

src/python/zquantum/core/measurement.py
```diff
@@ -347,31 +347,47 @@ def get_parities_from_measurements(
...
     values = []
     for _, term in enumerate(ising_operator.terms):
-        values.append([0, 0])
         marked_qubits = [op[0] for op in term]
-        for bitstring, count in bitstring_frequencies.items():
-            if check_parity(bitstring, marked_qubits):
-                values[-1][0] += count
-            else:
-                values[-1][1] += count
+        parity = check_parity_of_vector([*bitstring_frequencies.keys()], marked_qubits)
```
mstechly, line 358: [suggestion]I think it would be worthwhile to extract and call separately the method that transforms bitstrings into an array.

src/python/zquantum/core/measurement.py
```diff
@@ -412,48 +428,69 @@ def convert_bitstring_to_int(bitstring: Sequence[int]) -> int:
...
-    result = True
-    for qubit_index in marked_qubits:
-        if bitstring[qubit_index] == "1" or bitstring[qubit_index] == 1:
-            result = not result
-    return result
+    if not marked_qubits:
+        return np.ones(len(bitstrings_list))
+
+    # Convert bitstrings_list to np array
+    if type(bitstrings_list[0]) is str:
```
mstechly, line 449: [request] As mentioned elsewhere, this should be moved outside of the body of this function.

tests/zquantum/core/measurement_test.py
```diff
@@ -223,28 +223,26 @@ def test_convert_bitstring_to_int():
     assert convert_bitstring_to_int(bitstring) == 42
 
 
-def test_check_parity_odd_string():
-    bitstring = "01001"
+def test_check_parity_of_strings():
```
mstechly, line 226: Since `check_parity` should stay, these should be added as new tests, not replace old ones.

mstechly (changes requested):

src/python/zquantum/core/measurement.py
```diff
@@ -435,28 +446,69 @@ def check_parity(
...
+    bitstring_1d_array = np.frombuffer(all_bits.encode("utf-8"), "u1") - ord("0")
+    return bitstring_1d_array.astype(int).reshape(-1, n_qubits)
+
+
 def get_expectation_value_from_frequencies(
     marked_qubits: Iterable[int], bitstring_frequencies: Dict[str, int]
 ) -> float:
-    """Get the expectation value the product of Z operators on selected qubits
-    from bitstring frequencies.
+    """Get the expectation values of a set of operators (with precisions) from a set of
```
mstechly, line 484: [request] I think this docstring has been edited by accident. Could you revert the changes here?

src/python/zquantum/core/measurement.py
```diff
@@ -435,28 +446,69 @@ def check_parity(
     return result
 
 
+def check_parity_of_vector(
+    bitstrings_vector: np.ndarray, marked_qubits: Iterable[int]
+) -> np.ndarray:
+    """Determine if the marked qubits have even parity for each bitstring in the given
+        vector.
```
mstechly, line 453: I would add a comment here that it's similar in functionality to `check_parity` but much faster as it makes use of vectorization.

mstechly (changes requested):

src/python/zquantum/core/measurement.py
laurgao: Fixed in [df4fd65](https://github.com/zapatacomputing/z-quantum-core/pull/452/commits/df4fd654ebd2a866f21c8f0352caabf79c1ec68c)

src/python/zquantum/core/measurement.py
laurgao: Fixed in [19de1de](https://github.com/zapatacomputing/z-quantum-core/pull/452/commits/19de1de8df585a2f5117ce632456e343a6467a4d)

mstechly (approved):

PR #453, opened 2022-01-10
check validity of bitstring against np.integer type not native int type

MSRudolph:
If a bitstring is of type Tuple[np.int32, ...], the `is_measurement_outcome_distribution` check at initialization of `BitstringDistribution` will fail because `_are_keys_non_negative_integer_tuples` returns `False`. The reason is that `numpy.integer` instances are not instances of `int`.
See https://stackoverflow.com/a/37727662


mstechly (commented):
Looks good to me, but it falls under "there might be some weird numpy behaviour I don't understand" category, so I'll defer this one to @dexter2206 .

dexter2206 (approved):
LGTM

PR #455, opened 2022-01-11
ZQS-809 Updated circuit serde to support parameters with brackets

max-radin:


mstechly (approved):
Looks good to me, but wouldn't mind a second opinion from @dexter2206 or @alexjuda :)

src/python/zquantum/core/circuits/_serde.py
```diff
@@ -17,8 +18,21 @@ def serialize_expr(expr: sympy.Expr):
...
 
 
-def _make_symbols_map(symbol_names):
-    return {name: sympy.Symbol(name) for name in symbol_names}
+def _make_symbols_map(
+    symbol_names: Iterable[str],
+) -> Dict[str, Union[sympy.Symbol, Dict[int, sympy.Symbol]]]:
+    symbols_map: Dict[str, Union[sympy.Symbol, Dict[int, sympy.Symbol]]] = {}
+    for name in symbol_names:
+        # Check if the symbol name has brackets, such as "x[4]"
```
alexjuda, line 26: **suggestion**
It would be nice to add a comment stating the reason we look for brackets, WDYT about sth like:
```suggestion
        # Check if the symbol name has brackets, such as "x[4]". This is a workaround
        # for importing circuits from Qiskit v0.28 when the circuit is parametrized with
        # symbols from a "ParameterVector".
```
?

tests/zquantum/core/circuits/_serde_test.py
```diff
@@ -212,3 +219,11 @@ def test_load_save_circuitset_roundtrip(self):
...
         buf.seek(0)
         assert load_circuitset(buf) == circuitset
+
+
+def test_make_symbols_map():
+    symbols_map = _make_symbols_map(["x", "y[2]", "y[3]"])
+    assert symbols_map == {
+        "x": sympy.Symbol("x"),
+        "y": {2: sympy.Symbol("y[2]"), 3: sympy.Symbol("y[3]")},
+    }
```
alexjuda, line 229: **question**
`_make_symbols_map()` is used to generate `locals=` argument to [sympy.sympify](https://docs.sympy.org/latest/modules/core.html). Are we sure/was it tested that sympify() works with nested dictionaries?

`_make_symbols_map()` is a private function. WDYT about writing a unit test for `deserialize_expr()` instead? This would make the answer to my question above clearer + we we would only touch non-private stuff.

alexjuda (commented):
I've added a few suggestions :)

src/python/zquantum/core/circuits/_serde.py
```diff
@@ -17,8 +18,21 @@ def serialize_expr(expr: sympy.Expr):
...
+        match = re.search(r"^(.*)\[([0-9]+)\]$", name)
+        if match:
+            if symbols_map.get(match.group(1)) is None:
+                symbols_map[match.group(1)] = {}
+            symbols_map[match.group(1)][int(match.group(2))] = sympy.Symbol(name)
```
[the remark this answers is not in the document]
max-radin: neat!

src/python/zquantum/core/circuits/_serde.py
max-radin: Added similar comment in f32a85b80bb929fe6087ca3e510c236d81484b59

tests/zquantum/core/circuits/_serde_test.py
max-radin: Yup, it is tested indirectly through a [new item](https://github.com/zapatacomputing/z-quantum-core/blob/6eb5b957d400494af29fd26d47ad93f6b10ac8dd/tests/zquantum/core/circuits/_serde_test.py#L123-L127) added to `EXAMPLE_CIRCUITS` in the serde test.

Do you think that will be sufficient? Also I just deleted the test for _make_symbols_map in 84f6a684646d3b4042d8c6e1837e6c089f253f8d

alexjuda: Oh, right. Makes sense :)

alexjuda (approved):
LGTM!

PR #459, opened 2022-01-20
ZQS-925 fixing mypy configuration

max-radin:


src/python/zquantum/core/history/recorder.py
```diff
@@ -1,6 +1,6 @@
 """Main implementation of the recorder."""
 import copy
-from typing import Any, Callable, Dict, Generic, List, NamedTuple, TypeVar
+from typing import Any, Callable, Dict, Generic, List, NamedTuple, TypeVar, Union
```
mstechly, line 3: As far as I can tell this `Union` is unnecessary here.
```suggestion
from typing import Any, Callable, Dict, Generic, List, NamedTuple, TypeVar
```

mstechly (approved):
One minor comment otherwise looks good.

src/python/zquantum/core/history/recorder.py
max-radin: fixed in c7f962badf69175fbfdcfca8cb5d226bd5bbd78b

PR #462, opened 2022-02-01
fix(wavefunction): update  flip_amplitudes argument to StateVector

AthenaCaesura:
Had to increase the scope of flip_amplitudes so that it can deal with the new StateVector type.


src/python/zquantum/core/interfaces/backend.py
```diff
@@ -19,7 +18,7 @@
 
 # Note that in particular Wavefunction is a StateVector. However, for performance
 # reasons QuantumSimulator uses numpy arrays internally.
-StateVector = Union[Sequence[Complex], np.ndarray]
+StateVector = Union[Sequence[complex], np.ndarray]
```
mstechly, line 21: [suggestion] Perhaps move this to `zquantum.core.typing`?

src/python/zquantum/core/wavefunction.py
```diff
@@ -208,7 +208,7 @@ def flip_wavefunction(wavefunction: Wavefunction):
     return Wavefunction(flip_amplitudes(wavefunction.amplitudes))
 
 
-def flip_amplitudes(amplitudes: np.ndarray) -> np.ndarray:
+def flip_amplitudes(amplitudes: Union[Sequence[complex], np.ndarray]) -> np.ndarray:
```
mstechly, line 211: [suggestion] Why not use `StateVector` here?

mstechly (approved):

src/python/zquantum/core/interfaces/backend.py
AthenaCaesura: Konrad and I  have that planned for a different story.

AthenaCaesura: It's a different story because if we do that then we need to update the placement of `StateVector` in all the other repos.

src/python/zquantum/core/wavefunction.py
AthenaCaesura: Because importing `StateVector` from `backend.py` causes circular imports because `backend.py`needs to import from `wavefunction.py`. This will be fixed when we move `StateVector` to `typing.py`.

PR #470, opened 2022-02-18
Fix/zqs 936/athena caesura/python3.9.7 compatibility 2

AthenaCaesura:
Reject installs with python 3.9.7 to avoid bug with Protocol class.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import os
 import site
 import sys
 import warnings
+from platform import python_version
 
 import setuptools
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -43,6 +44,8 @@ def _read_readme():
         "Programming Language :: Python :: 3.7",
         "Operating System :: OS Independent",
     ],
+    # Avoid bug in Protocol in python 3.9.7
```
mstechly, line 47: I'd just add this comment for clarity, WDYT?
```suggestion
    # Avoid bug in Protocol in python 3.9.7. Note that higher and lower versions of Python are fine.
```

mstechly (approved):

PR #472, opened 2022-03-29
Faster custom gates

MSRudolph:
After discussing my use case with @AthenaCaesura, she suggested I try using my custom operations with `numpy.ndarrays` rather than `sympy.Matrix`. This alone reduces the runtime of my specific algorithm by an order of magnitude. I was able to cut the remaining run time more than in half by implementing a quicker bit flip operation. These changes made my application about ~40x faster.


AthenaCaesura (approved):

src/python/zquantum/core/circuits/_gates.py
```diff
@@ -353,9 +353,12 @@ def params_ordering(self) -> Tuple[Parameter, ...]:
...
     def __call__(self, *gate_params):
-        return self.matrix.subs(
-            {symbol: arg for symbol, arg in zip(self.params_ordering, gate_params)}
-        )
+        if isinstance(self.matrix, sympy.Matrix):
+            return self.matrix.subs(
+                {symbol: arg for symbol, arg in zip(self.params_ordering, gate_params)}
+            )
+        else:
+            return self.matrix
```
dexter2206, line 361: Line 361 is not covered. I suggest adding a test case for using `CustomGateMatrixFactory` with numpy array in place of sympy matrix.

dexter2206 (changes requested):
Looks good, but there is a test case missing for one of the added features.

AthenaCaesura:
Closing this PR as we have already moved over the the code from `z-quantum-core` to `orquestra-quantum` so it doesn't mkae sense to keep adding features to core. However, I am not closing this branch as it will be helpful in the future when making this update to `orquestra-quantum`.

PR #473, opened 2022-03-29
Getting rid of openfermion

Shiro-Raven:
In this PR, we include a slimmed-down version of openfermion in our core library to simplify our dependency graph which was mainly complicated by the transient dependencies of Openfermion. The slimmed-down version includes only the classes and functions needed by our libraries.

We also adapt the imports in core to the new configuration, and similar PRs are/will be created in other repos where openfermion is used. 

A known issue was the inability to install `core` and `runtime` together with pip 21. I re-ran the experiment using pip 22 and no problems were encountered.


src/python/zquantum/core/utils.py
```diff
@@ -607,34 +606,6 @@ def scale_and_discretize(values: Iterable[float], total: int) -> List[int]:
     return result
 
 
-def hf_rdm(n_alpha: int, n_beta: int, n_orbitals: int) -> InteractionRDM:
```
mstechly, line 610: Why removing this?

mstechly (approved):

src/python/zquantum/core/utils.py
Shiro-Raven: moved to `_utils.py` in `openfermion/zapata_utils` since it's more appropriate there

PR #474, opened 2022-03-30
Added function to add single qubit gate to several qubits

Shiro-Raven:


AthenaCaesura (changes requested):
Looks good, but you may need to bump the version of black to pass style. We can probably talk about this in the retro or sprint plannings today. https://github.com/psf/black/issues/2964

src/python/zquantum/core/circuits/_generators.py
```diff
@@ -17,28 +17,53 @@ def create_layer_of_gates(
...
+    qubit_indices: Iterable[int],
+    gate_factory: Union[GatePrototype, Gate],
+    parameters: Optional[np.ndarray] = None,
+) -> Circuit:
+    """Apply the passed gate to the specified qubits
+
+    Args:
+        circuit: circuit to add the gate to
+        qubit_indices: list of qubits that to apply the gate on,
+            duplicates will be ignored
```
dexter2206, line 44: Ignoring duplicates seem to be reasonable, however, users might still accidentally pass duplicates.

Hence, I would suggest raising an exception, or at the very least a warning if duplicates are present.

dexter2206 (changes requested):

src/python/zquantum/core/circuits/_generators.py
Shiro-Raven: Done! :)

AthenaCaesura (approved):

dexter2206 (approved):

PR #478, opened 2022-04-11
Distribution distance measures

SalonikResch5:
Total variation distance added as a distance metric for measurement outcome distributions


src/python/zquantum/core/distribution/moment_based_distance.py
```diff
@@ -0,0 +1,55 @@
...
+
+if TYPE_CHECKING:
+    from zquantum.core.distribution import MeasurementOutcomeDistribution
+
+
+def compute_moment_based_distance(
+    target_distribution: "MeasurementOutcomeDistribution",
+    measured_distribution: "MeasurementOutcomeDistribution"
+) -> float:
+    """Compute the total variation distance between two distributions,
```
AthenaCaesura, line 12: Is this comment the same as in the TVD? This is important as "moment based distance" is not quickly googleable. So we want to be precise here.

src/python/zquantum/core/distribution/moment_based_distance.py
```diff
@@ -0,0 +1,55 @@
...
+    Args:
+        target_distribution: The target probability distribution.
+        measured_distribution: The measured probability distribution.
+
+    Returns:
+        The value of the the total variation distance
+    """
+
+    #The starting value is 0
+    value = 0.0
```
AthenaCaesura, line 24: Would it be more intuitive to call this variable 'distance'? 🤔

AthenaCaesura (commented):
Are we sure that this function calculates the total variation distance? I'm certainly no expert here, but at a glance this doesn't seem to fit the [wikipedia definition of the TVD](https://en.wikipedia.org/wiki/Total_variation_distance_of_probability_measures).

You may also need to make another commit to rerun the tests properly. The github runner is kind of wierd with that.

PR #481, opened 2022-04-19
removing lea and redundant matrices from utils

Shiro-Raven:
I opted for applying the semantic of the function `sample_from_probability_distribution` instead of replicating the logic of the `lea` function


AthenaCaesura (approved):
Good! Much more readable than using lea!

PR #486, opened 2022-04-29
Unpacking breaks for usual parameter vectors

MSRudolph:
By looping `for parameter in parameters`, `parameter` can be a float. The current implementation breaks the function.


Shiro-Raven (approved):

mstechly (changes requested):
Could you add a test which checks this behaviour?
Also, I think it might not work if the parameters are sympy symbols, could you also check for  that?

AthenaCaesura (changes requested):
Also, shouldn't mypy take care of this? If `parameter` isn't an array (at least 1d) then you should get a mypy error.

MSRudolph:
`mypy` doesn't notice, because `parameters: np.ndarray` doesn't have a specified dimensionality. Try this:
```
from zquantum.core.circuits import Circuit, RX, apply_gate_to_qubits
apply_gate_to_qubits(Circuit(), (0, 1), RX, np.array([1, 2]))
```

PR #490, opened 2022-05-20
Convert Python version to 3.8

pcarinhas:
For Jira: ZQS-1017

* Change Python 3.7 to 3.8
* Add Makefile to allow local build test

### DIFF
--- a/docker/z-quantum-default/Dockerfile
+++ b/docker/z-quantum-default/Dockerfile
@@ -1,15 +1,18 @@
 # Dockerfile for the default z-quantum-core docker image
-FROM ubuntu
+FROM ubuntu:focal
 WORKDIR /app
 USER root
-RUN apt-get clean && apt-get update
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=Etc/UTC
+RUN ln -snf /usr/share/zoneinfo/$CONTAINER_TIMEZONE /etc/localtime && echo $CONTAINER_TIMEZONE > /etc/timezone
 
-# Install Python 3.7
+# Install Python 3.8
+RUN apt-get clean && apt-get update
 RUN apt-get install -y software-properties-common && \
     add-apt-repository -y ppa:deadsnakes/ppa && \
-    apt-get install -y python3.7 && \
+    apt-get install -y python3.8 && \
     apt-get install -y python3-pip && \
-    apt-get install -y python3.7-dev
+    apt-get install -y python3.8-dev
 
 RUN apt-get -y install \
                 wget \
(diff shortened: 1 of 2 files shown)


docker/z-quantum-default/Dockerfile
```diff
@@ -23,19 +26,22 @@ RUN apt-get -y install \
...
 
-# ZQuantum requires Python 3.7. By installing `python3-pip` we have also installed 
-# Python 3.8. We need to set back the Python version to 3.7.
-RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1 && \
-    update-alternatives --set python3 /usr/bin/python3.7
+#- # After March 2022, git changed its authentication.
+#- RUN git config --global url."git@github.com:".insteadOf git://github.com/
 
-ENV PYTHONPATH="/usr/local/lib/python3.7/dist-packages:${PYTHONPATH}"
+# ZQuantum requires Python 3.8. By installing `python3-pip` we have also installed
```
emre-aydin, line 36: Is there a typo here? Do you mean that `python3-pip` installed Python 3.9 or the comment is obsolete and needs to be deleted?

docker/z-quantum-default/Dockerfile
```diff
@@ -23,19 +26,22 @@ RUN apt-get -y install \
...
-                liblapack-dev
+                liblapack-dev \
+                openssh-client
 
 
-# ZQuantum requires Python 3.7. By installing `python3-pip` we have also installed 
-# Python 3.8. We need to set back the Python version to 3.7.
-RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1 && \
-    update-alternatives --set python3 /usr/bin/python3.7
+#- # After March 2022, git changed its authentication.
```
emre-aydin, line 33: I couldn't understand what this comment is about.

mesutcelik:
@zapatacomputing/platform When the PR merged, Does it mean that all clusters will pick that image up immediately with new submitted workflows. Can we instead publish new image with a new tag and give people opportunity with `custom-image` option of Orquestra Platform so that we set it default when we make sure it does not break anything.

ajroeth205:
> @zapatacomputing/platform When the PR merged, Does it mean that all clusters will pick that image up immediately with new submitted workflows. Can we instead publish new image with a new tag and give people opportunity with `custom-image` option of Orquestra Platform so that we set it default when we make sure it does not break anything.

Since this is being merged into the `dev` branch, the existing Concourse pipeline will build an image tagged `dev`. I thought QS team was no longer using the `dev` branch flow, but the `dev` branch is not behind `master`, so maybe that hasn't happened yet? (tagging @max-radin.) Also, [Athena has tested this image](https://zapatista.slack.com/archives/CBADQT75X/p1652906601504559?thread_ts=1652905671.559199&cid=CBADQT75X) using the `test_image` tag.

docker/z-quantum-default/Dockerfile
pcarinhas: Will remove..

docker/z-quantum-default/Dockerfile
pcarinhas: Will remove. It was about how the old dockerfile was trying to clone. Its not really helpful now.

max-radin:
Merging to `dev` should not change the default behavior of Quantum Engine since a push to `latest` is only triggered on merge to `main`. The new repos such as `orquestra-quantum` use github flow (meaning there's no dev branch). However we didn't change `z-quantum-core` to github flow because we weren't planning to do further releases.

pcarinhas:
> Merging to `dev` should not change the default behavior of Quantum Engine since a push to `latest` is only triggered on merge to `main`. The new repos such as `orquestra-quantum` use github flow (meaning there's no dev branch). However we didn't change `z-quantum-core` to github flow because we weren't planning to do much further development.

   Would it be ok then to remove the dev branch in the near term on this repo?

max-radin:
Switching this repo to github flow would require a bit more than removing the dev branch: we'd also have to update the github action workflows in this repo as well as the release script for the z-quantum-* repos. Which we could do, but I don't think that this would make it any easier to transition other repos.

ajroeth205:
Is the `dev` branch of `z-quantum-core` broken? I notice the image build pipeline for it is red. If so, I think this PR should be off of a working version of the image instead (even if we want to move it to `orquestra-core`) to make sure it's modifying the most current version of the image instead of a broken one.

max-radin (approved):
@ajroeth205 I think in this case it's probably not worthwhile to separately fix the failure with the current `dev` dockerfile before merging this PR. It looks like most likely that failure was not due to a change in our dockerfile, but rather because of external changes. (For example we did not pin the ubuntu version, so new ubuntu releases could have broken our image build.)

Also a note here about the build process: I think it's safe to merge this into `dev` now since it will not trigger a push to `latest`. (To push to `latest`, we would need to both unpause the `z-quantum-default-main` pipeline and merge these changes into `main`.)

emre-aydin (approved):

ajroeth205:
> @ajroeth205 I think in this case it's probably not worthwhile to separately fix the failure with the current `dev` dockerfile before merging this PR. It looks like most likely that failure was not due to a change in our dockerfile, but rather because of external changes. (For example we did not pin the ubuntu version, so new ubuntu releases could have broken our image build.)
> 
> Also a note here about the build process: I think it's safe to merge this into `dev` now since it will not trigger a push to `latest`. (To push to `latest`, we would need to both unpause the `z-quantum-default-main` pipeline and merge these changes into `main`.)

Okay, then I will temporarily set the dev pipeline to build the tag `0.20.0`. @max-radin @pcarinhas

PR #495, opened 2022-07-25
Bugfix parameter unpacking in gate_factory

MSRudolph:
The function would break if the passed `parameters` are a 1d array. `parameter` would then be a float.

@roberCO found this bug in `z-quantum-core`. I thought I once made a PR about this, but apparently it was never merged.

### DIFF
--- a/src/python/zquantum/core/circuits/_generators.py
+++ b/src/python/zquantum/core/circuits/_generators.py
@@ -61,7 +61,7 @@ def apply_gate_to_qubits(
         assert len(parameters) == len(unique_qubit_idx)
         gate_factory = cast(GatePrototype, gate_factory)
         for qubit, parameter in zip(unique_qubit_idx, parameters):
-            circuit += gate_factory(*parameter)(qubit)
+            circuit += gate_factory(*np.atleast_1d(parameter))(qubit)
     else:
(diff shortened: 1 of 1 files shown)


AthenaCaesura:
Remember that we aren't doing releases anymore on `z-quantum-core` so this will never reach the `main` branch. We should probably focus on putting this change into `orquestra-quantum` instead. I'll approve it for now because there's no good reason not to. But we really are pushing for people to move over to the new `orquestra-*` libraries.

AthenaCaesura (approved):

MSRudolph:
Oh wait, I honestly thought I was creating this PR for `orquestra-core`. Sorry @AthenaCaesura!

AthenaCaesura:
Np! I already made the PR to orquestra-quantum. 😄 Sorry your previous PR didn't get merged here. We do our best but sometimes things slip through the cracks....

MSRudolph:
> Np! I already made the PR to orquestra-quantum. 😄 Sorry your previous PR didn't get merged here. We do our best but sometimes things slip through the cracks....

Thanks :) 
Should we still merge this to `main`?

mstechly:
It should be to `dev`, not `main` - I changed it in the PR.
We could merge it to `dev`, but I'd rather not if not necessary.

